PR #31, opened 2016-03-19
qa-escapes

gunnarbittersmann:
updated German translation


gunnarbittersmann:
l will open another pull request without commits from other branches.

PR #32, opened 2016-03-25
Translation contributor credits

gunnarbittersmann:
Introducing f.translationContributors as optional field for translators to give credit where credit is due.
Its label is s.translationAcknowledgements which needs to be included in the boilerplates.
(This could be done together with s.acknowledgements which is still missing in some boilerplates.)


r12a:
i have reservations about introducing this, as mentioned in email. Not least is that it requires me to obtain additional translations for all languages.  As it's not a critical issue, i will come back to it some time in the future.

xfq:
Thank you for the suggestion. My inclination is to use `f.contributors` to acknowledge the efforts of the translation contributors.

Closing.

PR #37, opened 2016-04-20
Replace "girl" with "woman"

jkshapiro:
Usually children don't get married.


r12a:
Thank you, but usually 'girl' doesn't just refer to children. See for example, the definition in the merriam-webster online dictionary that includes the definitions "a young unmarried woman", and "sweetheart".

PR #62, opened 2016-09-07
Fix broken links

hail2u:
The link that references `#answer` in “Declaring character encodings in
HTML” is broken. This should reference `#quickanswer`.


r12a:
thanks for pointing this out, and providing a PR !   Actually, that link is supposed to refer to the Details section rather than the Quick Answer section.  I have made those changes at https://github.com/w3c/i18n-drafts/commit/a3341cb022a745029de6aef49c4f36fca5c88273 so i'll just close this PR now.

PR #71, opened 2016-12-08
Fixing 2 lang attributes

notabene:
A parasitic sv lang attribute made its way into the final file, don't know from where it came. Also, I had forgotten a lang attribute on an English segment, ironically!


r12a:
Thanks Stephane.

PR #85, opened 2017-04-30
Minor text edits.

aphillips:


aphillips:
Hi Richard, I don't have permissions on i18n-drafts. Can you do the honors?

PR #194, opened 2019-06-12
Spanish name correction

cmatmar1:
Hello Richard!

I found a mistake on the name María-Jose Carreño Quiñones, as José would be written with the diacritical mark and separated from María, being María José the correct form of that name without the hyphen. I also added in brackets that José is the second given name of María, as many Spanish-speaking people have two given names.

Kind regards,

Carlos Mateo


r12a:
Thanks, Carlos.

r12a:
Note that i subsequently changed the text in brackets to 

> (José being a part of her given name)

For the record, i also changed the French, German and Ukrainian versions.

PR #197, opened 2019-07-09
update questions/qa-headers-charset.fr.html

brandelune:
Let me know if there is something I missed.


brandelune:
I'm not sure where the conflicts come from... :(

r12a:
@brandelune your PR seems to include commits for another file, and those for the Vérifier translation are trying to overwrite an old version.  Did you use an existing branch, maybe, to submit the PR, rather than creating a new one from the latest version of the repo?

brandelune:
I guess there is something like that. I'll do that again.

PR #200, opened 2019-07-14
Created tutorial-char-enc/index.fr.html

brandelune:


r12a:
LGTM, thanks!

Should i publish+announce it now, or should we batch it up with one or more upcoming translations?

brandelune:
Just do as you wish, no rush here.

brandelune:
Just in case you forgot, the article is still not published. I had forgotten that you had delayed publication and was confused this morning when I saw the site did not have the FR version... :)

r12a:
I've been waiting for more to come in, so that i can announce them all together (see that the status is 'Received' at https://github.com/w3c/i18n-drafts/projects/2, rather than 'Published').  But maybe i should just publish, to avoid confusion.

r12a:
It's now published. Thanks.

Issue #208, opened 2019-07-22
Translated boilerplate needs update


r12a:
The following boilerplate text was changed in a recent update, but not for all languages.

> s.aboutI18nActivity = "Get help from W3C Internationalization."
> s.about = "Contact"
> s.groupsThatMakeUp = "Participate in W3C Internationalization work."
> s.groups = "Participate"
> s.topicIndexForInformation = "Search this site."
> s.topics = "Search"
> s.taskBasedIndex = "Learn techniques and requirements for internationalization."
> s.techniques = "Learn"
> s.informationResources = "Find information on this site."
> s.resources = "Find"
> s.newsFiltersAndFeeds = "Follow the work at W3C Internationalization."
> s.news = "Follow"

These items of text appear just below the word 'Internationalization' at the top of the page. The one-word items are the visible text, and the longer items are the title attribute text behind them.

> s.lastChanged = "Last changed "

The string above appears at the very bottom of the article list, and is followed by a date+time such as 2019-07-23 12:30.

Retranslations are needed for:
~~ar,~~ ~~bg,~~ ~~de,~~ el, ~~es,~~ ~~fr,~~ ~~hi,~~ it, hu, ~~ja,~~ ~~ko, nl~~, pl, pt-br, pt, sv, tr, uk, ~~zh-hans,~~ zh-hant

(Some languages need translations for additional strings, too.)

r12a:
@brandelune would you be able to provide the French translations?

brandelune:
Where does that appear ?

brandelune:
Sure !

r12a:
Ah, i had meant to add that. Thanks for the reminder. 

These items of text appear just below the word 'Internationalization' at the top of the page. The one-word items are the visible text, and the longer items are the title attribute text behind them.

brandelune:
Ok, I'll be on it tomorrow.

xfq:
zh-hans updated: https://github.com/w3c/i18n-drafts/pull/209

(BTW, some property names do not look very semantic to me.)

r12a:
> (BTW, some property names do not look very semantic to me.)
which ones?

ntounsi:
I can update (create) for Arabic.

r12a:
That would be great, @ntounsi, thank you. I'll send you an email, since there are some additional things we can do for Arabic.

ntounsi:
Sure!

brandelune:
[...]

PR #210, opened 2019-07-23
Update boilerplate text of ko-kr

techhtml:
#208


r12a:
Many thanks @techhtml !

r12a:
@techhtml there are a few more strings in the document that need translation, and also the page at https://www.w3.org/International/articlelist.ko (basically just intro and headings).  Please let me know if you are interested in translating those too.  (the latter file isn't on github, so i'll need to send you a file.)

techhtml:
@r12a Sounds Great! I will ping you when I find some interested docs.

r12a:
Sorry for not being clear, @techhtml, but i meant that we need [1] some more strings translated in the file https://github.com/w3c/i18n-drafts/blob/gh-pages/javascript/boilerplate-text/boilerplate-ko.js (especially near the bottom), and [2] the actual page at https://www.w3.org/International/articlelist.ko

Of course, it would be great to also get translations of the articles listed on that page too, but our immediate need is to address the shortfall for the boilerplate text.  Hope that's clearer.

techhtml:
@r12a a-ha, Sorry for my misunderstand haha.
I will translate this docs Korean and will open PR soon.

PR #212, opened 2019-07-24
updated translation to 3d7a13e (for #208)

himorin:
for #208 
going through all lines as 3d7a13e (fixing some in addition to updates pointed in #208 )


r12a:
Many thanks, @himorin !

PR #213, opened 2019-07-25
translated article-list into Japanese

himorin:
per comment https://github.com/w3c/i18n-drafts/issues/208#issuecomment-514587317, and request via email.


r12a:
Thanks @himorin  !

PR #214, opened 2019-07-25
Update boilerplate-ar.js

ntounsi:
Arabic translation done. Needed preview to fix bidi stuffs.

### DIFF
--- a/javascript/boilerplate-text/boilerplate-ar.js
+++ b/javascript/boilerplate-text/boilerplate-ar.js
@@ -43,68 +43,67 @@ s.ltrAttribute = " dir='ltr' "
 s.rlm = "‏"
 
 
-
 // TRANSLATE THE FOLLOWING
 s.i18nActivityHome = "الصفحة الرئيسية لنشاط I18N"
 s.moreResourcesOfThisType = "المزيد من هذا النوع من الموارد."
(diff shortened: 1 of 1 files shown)


r12a:
Thanks very much, Najib!

r12a:
Ah, i just spotted one that you missed:

s.followOurNews = "Follow our news feed."

Could you provide the string here (rather than create a new PR), and i'll insert it.

Also, would you be able to translate the small amount of text in https://github.com/w3c/i18n-drafts/blob/gh-pages/article-list/articlelist.ar.html ?

ntounsi:
On 7/25/19 11:53, r12a wrote:

> Ah, i just spotted one that you missed:
>
> s.followOurNews = "Follow our news feed."
>
> Could you provide the string here (rather than create a new PR), and 
> i'll insert it.
>
تابع موجز أخبارنا

meaning : "Follow our news summary". Another translation would be:

تابع مجرى أخبارنا

which means : "Follow our news stream".


> Also, would you be able to translate the small amount of text in 
> https://github.com/w3c/i18n-drafts/blob/gh-pages/article-list/articlelist.ar.html 
> ?
>
Yes.

r12a:
Done. Thanks.

PR #218, opened 2019-07-27
Create article-text-size.fr.html

brandelune:


brandelune:
There must be a way to fix my previous mess that does not involve deleting my fork :)

r12a:
This has now been published and announced. Thanks !

PR #220, opened 2019-08-04
Create index.fr.html

brandelune:
There may be an issue with the length of the svg strings. Can you check that on your side ?


r12a:
Sorry for the hiatus, due to Geek Week intervening. I fixed the issue with the string length, and also converted 3 more images to SVG (and translated the text). I also uploaded the files to the W3C server and github.  Thanks !

brandelune:
Thank you !

PR #233, opened 2019-12-20
switched w3c_home from http to https (mixed contents)

himorin:
for loading w3c.github.io/i18n-activity/ as https, image loading from http causes mixed contents warning and will be blocked on Chrome shortly.
This is to fix w3c_home icon loaded from https

for(?) https://github.com/w3c/i18n-activity/pull/830


r12a:
Thanks @himorin

Issue #255, opened 2020-09-02
[questions/qa-personal-names]  more on titles?


HedvigS:
[source] (https://www.w3.org/International/questions/qa-personal-names) [en]

I really, really liked this piece. It captures a lot of the problems I've encountered. Thank you for writing it.

If you think it's a good idea, I would consider adding some more on titles, and the lack thereof. I'm Swedish, and living abroad (Australia and Germany) I actually find it very frustrating to be forced to use titles such as Ms., Frau etc constantly. Where I'm from we don't use titles at all. Why does the bus company need to know if I'm married? Or why do I need to use "Ms." to explicitly tell them that I don't want them to know my marital status? Surely if I want the title field to be blank that should be okay?

It seems like the solution is a freeform text box. Even in very extensive dropdown menus it is impossible to include all possible types of titles the world over. But as you say in the piece, the first question should be: **do I really need this information?**

Something that ironically isn't often asked but is more relevant for many UI systems is pronouns. Whether you want "See her details" etc. If this is what they are hoping the titles question will get them, why not ask that directly?

HedvigS:
It might also avoid funny things like this :P
![117755507_10163964769965029_5896318070277162033_n](https://user-images.githubusercontent.com/5327845/91999796-10de1800-ed3d-11ea-9524-19df73c67cd0.jpg)
![117798113_10163964769970029_4463093033826951749_n](https://user-images.githubusercontent.com/5327845/91999799-120f4500-ed3d-11ea-88e4-49b39965fefc.jpg)

xfq:
It is mentioned a bit here: https://www.w3.org/International/questions/qa-personal-names#formality , but I agree that more information on honorifics/titles is useful.

HedvigS:
Yeah I saw that and that was nice. Just thought it could be expanded on a bit, in particular if form designers are just using it as a shortcut to figure out what pronoun to use.

r12a:
How's this: https://github.com/w3c/i18n-drafts/pull/256/files

r12a:
Closing this.  Reopen if there are additional comments, or comment on the PR. Tack själv!

PR #256, opened 2020-09-04
Add guideline about requiring titles

r12a:
Addresses the comment at https://github.com/w3c/i18n-drafts/issues/255


r12a:
I'm going to merge this, so that it doesn't create conflicts with other updates i want to add. But i can still incorporate any comments raised here after merging by re-editing the page.  This will only change the GH shadow of that page anyway.

Issue #98, opened 2017-07-10
[articles/definitions-time/index]  


wsdwl:
[source] (https://www.w3.org/International/articles/definitions-time/index) [en]

I like the definition and most sounds right to me.

My one remark would be that reducing the leap year rule just to the first part is wrong. You should either say, that a leap day is inserted usually every 4 years or include the full explanation with "except when the year is evenly divisible by 100, unless it’s also evenly divisible by 400".

r12a:
@aphillips what should we do here ?

aphillips:
@wsdwl is correct. Having played in this window with text proposals for a few minutes... I think defining the rule makes the most sense. You want a PR? Or just text:

> The Gregorian calendar is the most widely used way of representing civil time. It is a solar calendar, with years usually consisting of 365 days, plus the concept of a "leap year". This adds an additional day every 4 years, except when the year is evenly divisible by 100 (unless the year is also evenly divisible by 400).

I'm resisting the temptation to say that the calendar counts years in the Common Era.

r12a:
Thanks, both. Fixed in https://github.com/w3c/i18n-drafts/pull/259.

PR #259, opened 2020-09-07
Update index.en.html

r12a:


r12a:
Fixes https://github.com/w3c/i18n-drafts/issues/98#issuecomment-687246433 with wording from @aphillips .

PR #291, opened 2021-04-25
Example update: "If you don't know the direction of the phrase" section

fsasaki:
The example fits better into that section with the proposed update.


r12a:
Thanks for spotting that.  I used bdi rather than span (this was the ideal place to put the one example of that element).  So i committed my own version, and will close this.

PR #300, opened 2021-06-04
Increase contrast

xfq:
Make the links in the sidebar clearer.

### DIFF
--- a/style/article-2016.css
+++ b/style/article-2016.css
@@ -233,7 +233,7 @@ aside li {
 
 a {
 	text-decoration: none;
-	color: teal;
+	color: #007575;
 	}
 #toc a {
     color: darkslategray;


r12a:
@xfq it's not clear to me which links you have a problem with in the sidebar.  Sidebar links on a beige background should be a slate grey colour.  Do you mean links in side notes?  Changing that line in the CSS doesn't seem to produce a difference for me.   Perhaps you could post here a picture of what you're wanting to change.

xfq:
I'm talking about the breadcrumb navigation:

links

It may be because of the color difference. I can see them clearly on one monitor, but they are not so clear on another monitor.

r12a:
Thanks, i now understand which links you are concerned about.  

However, the `teal` and `#007575` colours seem the same to me. Could you check that?

xfq:
The two colours are similar, but not exactly the same ([#008080](https://drafts.csswg.org/css-color/#valdef-color-teal) and `#007575`). I tried not to change the original colour scheme as much as possible, so I didn't make major adjustments.

r12a:
Understood.  It's just that, when flipping between the 2 colours using the inspector, i see no difference at all to the contrast of the text.  I suppose that's a good thing, since it means there's no harm done, but if there's a difference on one of your systems i wonder what the differences are for others we haven't seen.  Anyway, i suppose we can go ahead.  

(One reason for me to hesitate is that it's much easier to remember and type 'teal' than `#007575`, but such is life. I'd be curious to see the same image you added above with the alternate colour, and to know whether you and i see the 2 images differently.  If you have a spare moment?)

xfq:
Looking at it again, I think the changes are not big enough, and we don’t use `style/article-2016.css` much now. So, closing.

PR #302, opened 2021-06-07
Initial draft of 'developing localizable manifests'

aphillips:
[I18N-ACTION-1033]


techniques/developing-localizable-manifests.en.html
```diff
@@ -0,0 +1,216 @@
+<!DOCTYPE html>
+<html dir="ltr" lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Developing Localizable Manifests</title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" async="" class="remove"></script>
+    <script class="remove">
+      var respecConfig = {
+          specStatus: "ED",
+	      edDraftURI: "https://w3c.github.io/i18n-drafts/techniques/developing-localizable-manifests.en.html",
```
xfq: Will this be an [article](https://www.w3.org/International/articlelist) or a document on /TR (like string-meta and specdev)?

techniques/developing-localizable-manifests.en.html
```diff
@@ -0,0 +1,216 @@
...
+
+    </section>
+    <section id="introduction">
+      <h2>Introduction</h2>
+
+        <p>Some specifications on the Web deal with defining sets of files or resources to be consumed together. A common design pattern is to provide a manifest or configuration file that defines how various resources should be used or to provide various kinds of metadata about a collection of resources.</p>
+        
+        <p>Some examples of specifications like this include Webapp [[APPMANIFEST]], Miniapp [[MINIAPP-MANIFEST]], and ePub [[EPUB-33]].</p>
+
+        <p>Manifest files generally contain a mix of <a href="https://w3c.github.io/i18n-glossary/#def_localizable_content">localizable content</a> and <a href="https://w3c.github.io/i18n-glossary/#def_syntactic_content">syntactic</a> or <a href="https://w3c.github.io/i18n-glossary/#def_user_supplied_value">user-supplied values</a> that are not localizable. A specification that defines such a format needs to provide a means for users to localize the manifest into different languages. This document covers the pros and cons of common designs, along with some best practices related to building and managing manifests.</p>
```
xfq: Would it be better to change "Manifest files generally" to something like "Some manifest files"?

For example, it seems that there is no human-readable text in [Payment Method Manifest](https://www.w3.org/TR/payment-method-manifest/) or the [origin policy manifest](https://wicg.github.io/origin-policy/#manifest-file).

xfq:
Do we want to add some examples of hybrid approaches?

FYI - this proposal for MiniApps Manifest looks like one such example: https://pr-preview.s3.amazonaws.com/MichaelWangzitao/miniapp-manifest/pull/12.html#i18n-solution-i18n

techniques/developing-localizable-manifests.en.html
aphillips: It's a good question. I used Respec with the idea that it would probably become a note, but I used the i18n-drafts repo instead of creating a new, separate, repo. The other "techniques" documents are actually indexes, so this doesn't really fit as well as I thought it might.

techniques/developing-localizable-manifests.en.html
aphillips: Good callout. See what you think of the updated text.

xfq:
BTW, we can use `<pre class="json">` to syntax highlight the content inside `pre` elements.

techniques/developing-localizable-manifests.en.html
r12a: This doesn't belong in the drafts/techniques directory, for sure.  It should probably have its own repo, so that we can publish it as a Note in the future.

Issue #306, opened 2021-07-14
Broken link


brandelune:
https://github.com/w3c/i18n-drafts/blob/94b9db54cf2c1fa41e19eda2c2f8e1b4f0def846/questions/qa-choosing-encodings.en.html#L118

The correct URL is:
https://www.w3.org/TR/html50/document-metadata.html#charset

brandelune:
Sorry, gh-pages, not 94b9db5.

PR #307, opened 2021-07-14
broken link to HTML spec

brandelune:
cf. issue #306


r12a:
Thanks for this PR.  As it happens, i was quick off the mark to fix this, so they changes are made.  (I had to make the changes in the 4 translated versions of the page, too, anyway.)

So i'll just close this.  Thanks again.

PR #314, opened 2021-09-27
qa-personal-names: Make partial changes global.

arkhi:
This is a follow‑up of af74c77113e80b91098e791f89556744df2e18c3.


r12a:
Many thanks for this fix, which seems to have slipped under the bridge unnoticed but which i just spotted again.

PR #320, opened 2021-11-03
Fix broken link

xfq:
The URI fragment to Authoring HTML & CSS in https://www.w3.org/International/questions/qa-personal-names.en#endlinks is broken.


r12a:
Thanks for spotting that, @xfq, but we're going to need to change it in all the translations too. (And then update the CVS copies.)  Would you be able to add an additional commit for the translations?

xfq:
I can update the link in the translations, but I don’t know how to translate the text of the link, so I didn't update them...

xfq:
Because that page is currently only available in English, I updated the PR to use the English name.

r12a (approved):

PR #324, opened 2021-11-18
Update links to BCP 47

### LINKED ISSUE #321: 404 for http://tools.ietf.org/html/bcp47
https://www.w3.org/International/questions/qa-choosing-language-tags includes the link [2.2.7 Private Use Subtags](http://tools.ietf.org/html/bcp47#section-2.2.7), which yields a 404.

xfq:
Fix https://github.com/w3c/i18n-drafts/issues/321

### DIFF
--- a/articles/language-tags/index.en.html
+++ b/articles/language-tags/index.en.html
@@ -246,9 +246,9 @@ Suppress-Script: Latn
       <pre class="tagExample" translate="no"><span class="languageExample currentExample" title="Language subtag">en</span></pre>
       <pre class="tagExample" translate="no"><span class="languageExample currentExample" title="Language subtag">ast</span></pre>
       <p class="readmore">Read more in the BCP 47 spec: </p>
-      <p><a class="print" href="http://tools.ietf.org/html/bcp47#section-2.2.1">2.2.1 Primary Language Subtag</a></p>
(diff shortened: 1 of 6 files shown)


r12a:
I see that you're using links like https://tools.ietf.org/search/bcp47#section-2.2.7 rather than https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.7.  Do you think that's better?

xfq:
I think these two links are similar to https://www.w3.org/TR/jlreq/ and https://www.w3.org/TR/2020/NOTE-jlreq-20200811/

The first one always links to the latest version of the IETF language tag standard, and the second one may be obsoleted by another RFC document. For those who want to link to a stable version, the latter is better, and for those who want to always point to the latest version, the former is better.

Another difference is that BCP 47 contains two RFC documents. I linked to BCP 47 here because [specdev recommends linking to BCP 47](https://w3c.github.io/bp-i18n-specdev/#lang_bcp_not_rfc).

r12a (approved):

xfq:
(Before merging this, I will discuss this issue with the i18n WG first.)

xfq:
Closing in favor of https://github.com/w3c/i18n-drafts/pull/329

PR #340, opened 2022-01-24
fix: switch (many) URLs to “https”

j9t:
(Note: IDE automatically removed trailing spaces—as this seems useful, I kept this change)

Signed-off-by: Jens Oliver Meiert <jens@meiert.com>


r12a:
@j9t first, thanks for raising this PR.

However, i'm a little concerned about applying it to such a large number of files.  This is because we have numerous articles undergoing updates or in the process of translation, and applying this change will produce conflicts when trying to merge those files.

If we could identify which files are stable and don't have changes pending, this would be great, but that will take a while.  So i'm inclined to rather make these changes each time we touch a file. 

Note that i also regularly change other things before updating a file or sending out for translation, including:
1. remove the shim
2. remove  type="text/javascript"
3. remove  type="text/css"

(Those may not have been done for the file you just translated, for which apologies.)

So i'll close this PR, unless you think i'm mistaken, in which case please reopen.

j9t:
@r12a, I see your point about conflicts, but these should be instances easily resolved automatically. I’d think that in practice, they lead to no conflicts at all.

Just my two cents, it’s your call of course.

PR #347, opened 2022-02-17
Try adding syntax highlighting

xfq:
This PR attempts to add syntax highlighting to our code to improve the readability. The currently supported languages are HTML/XML, CSS, JavaScript, HTTP, and JSON, which should basically meet our needs. It is also easy to extend if needed.

Here is a comparison:

before

after

(Specific colors can be customized if needed.)

-----

The code changes are few and simple. Just use the [recommended way to mark up a code block](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element) (a `<pre>` element with a `<code>` element inside), like so:

```html
<pre><code class="language-css">p { color: red }</code></pre>
```

It uses Prism, which is a lightweight syntax highlighter invented by Lea Verou.


xfq:
Merging.

PR #346, opened 2022-02-17
Mention sitepage_2021.js

xfq:

PR #348, opened 2022-02-24
Revert the changes to sitepage.js

xfq:
As I said in https://github.com/w3c/i18n-drafts/pull/346#issuecomment-1045522788 , since we have https://github.com/w3c/i18n-activity/commit/46f4df186b375bafbddeb2b72ec803ea11d1b899 , we can update that page in the future.


r12a:
I prefer to keep this note in the js file, because i won't normally go read the editorial guidelines before fixing bugs, etc.  So having this at the top is a useful reminder that i should be working on a different file.  I actually spent quite a lot of time prior to editing the guidelines last week trying to figure out which .js file was current and which one i should point to.  If i'd have a note like this at the beginning of the .js file, i'd have saved a great deal of time.

xfq:
Perhaps we can change this note to something like "see the latest version in the editorial guidelines". Otherwise, every time we create a new version of the js file, we need to change all the old versions to point them to the latest version.

r12a:
Actually, i prefer it as is because the old file is still in use and pointing to the immediate successor helps if you want to see what changed. The guidelines will  only show the latest, but remember that the old versions will be in use for a considerable time. If we have a new version in the future, we can simply chain link to the latest, if that's what we're looking for.

r12a:
> If we have a new version in the future, we can simply chain link to the latest, if that's what we're looking for.

ie. i'm not proposing that we change all .js files to point to the latest when we create a new version - each will point to the next only.  It's then trivial to find the latest, if that's what we need.  And when i say 'create a new version', i mean something that's not backwards compatible or where the name has changed and that needs to be reflected in new articles/pages.  We, of course, can still update a particular version using commits or PRs for less substantial changes.

xfq:
@r12a Fair enough. Closing this PR.

PR #357, opened 2022-04-27
Added code example for ModPerl 2

randolf:
Current text: https://www.w3.org/International/articles/http-charset/index.en.html#scripting


xfq:
@duerst Any objections to merging this?

duerst:
@xfq No, no objections. But I can't guarantee that it's correct, because I don't know ModPerl 2.

xfq:
Thank you, @randolf and @duerst .

Merging.

PR #361, opened 2022-05-05
Translate the about page into Chinese

xfq:
This PR:

* translates the 'About internationalization' page into Simplified Chinese
* add some unique styles for Chinese, including counters and emphasis marks
* fixes typos in the English version


nav/about-data/style.css
```diff
@@ -0,0 +1,11 @@
+.nowrap {
+    white-space: nowrap;
+}
+
+.em-zh {
```
r12a: Is there a reason we aren't using :lang(zh) em here ?

nav/about.zh-hans.html
```diff
@@ -0,0 +1,412 @@
...
+
+<p>从一开始，Web for All就一直是万维网联盟的设计原则和目的，也是对Web应用的自然要求，因为世界各地的人们都可以使用Web。遗憾的是，很多人很容易忽视与其所在文化不同的人或使用不同语言或书写系统的地域的需求，而这样构建的Web应用和内容会阻碍世界各地的很多用户使用您的技术或内容。</p>
+
+
+
+
+
+<section id="sidebarExtras">
+<h2 id="i18n" class="notoc"><a href="#i18n">什么是i18n?</a></h2>
+<p>i18n是internationalization（<span class="nowrap">国际化</span>）的<span class="nowrap">行业标准</span><span class="nowrap">缩写</span>（因为首末字符i和n之间有18个子母）。</p>
```
r12a: Are you happy for wrapping to leave a ( at the end of a line?  If not, you'd need to put the parens inside the span.

xfq:
We should add https://github.com/w3c/timezone to this page as well, but that is a separate issue.

r12a (changes requested):

nav/about-data/style.css
xfq: No, so I fixed this using `em:lang(zh)`.

I also removed some `em`s in the Chinese translation, because Chinese rarely adds emphasis marks to a long sentence/paragraph (like the descriptions of groups in https://www.w3.org/International/i18n-drafts/nav/about#groups ).

nav/about.zh-hans.html
xfq: Fixed.

r12a (approved):

PR #367, opened 2022-05-15
A number of small updates including the Getting Started page update

brandelune:


r12a:
@brandelune thank you for suggesting all these http->https changes, but unfortunately things are a little more complicated than they seem.  Most of these files are available via GitHub, but are actually published on the W3C site under CVS (for language negotiation to work).  So i would need to also check and change all 300 files in another location of my hard drive then upload them using the command line.  It's far too much work, so we've been making these fixes as we change files.  Your list suggests some files i can change, and i'll do those, but i can't merge this PR. Sorry.

r12a:
I suspect that you included the translation i was waiting for with the 299 or so other changes.  Could you submit that in a separate PR?  (It's always best to keep separate substantive changes and those that make large scale format or typo fixes, so that it's possible to see more clearly what is substantive.)

brandelune:
Ok. Let me cancel the PR. You know where the issue is, and it's not important anyway since http is automatically redirected to https.

So, I'll recreate a PR with only the modifications to the French files I made. Would that work better ?

I'll do a PR for the getting started alone, and a PR for the other things.

r12a:
> I'll do a PR for the getting started alone, and a PR for the other things.

That would be perfect. Thanks.

r12a:
@brandelune i just noticed that some of your edits add `type="text/javascript"` to a file.  Please don't add that to your translation.  In fact, we have been systematically removing that code from our pages, because it produces a warning in the HTML validator.  We are also removing `type="text/css"`.

brandelune:
There is definitely an issue on my side. Sorry for the hassle. I'll be checking the thing later this week.

PR #371, opened 2022-05-24
Add ids to the groups

xfq:
This PR adds ids to the groups in https://www.w3.org/International/i18n-drafts/nav/about#groups

If this looks good to you, I'll update the Chinese version as well, @r12a.


xfq:
Merging per discussion with @r12a offline.

PR #374, opened 2022-05-30
Translate qa-when-lang-neg into Chinese

xfq:
This PR:

* translates [When to use language negotiation](https://www.w3.org/International/questions/qa-when-lang-neg) into Simplified Chinese
* removes the space after colon in the Chinese template
* updates the link to the HTTP Cookie RFC
* fixes a few typos in the English version


javascript/doc-structure/article-2022.js
```diff
@@ -227,7 +227,11 @@ var modCredit = ''
 if (f.modifiers && f.modifiers != '') modCredit = s.modifiedBy+' '+f.modifiers+s.sentenceDelimiter
 
 var translatorCredit = ''
-if (g.isTranslation) translatorCredit = s.translatedBy+' '+f.translators+s.sentenceDelimiter
+if (g.isTranslation) {
+	// Remove the space in Chinese
```
r12a: An alternative approach would be to include the space in s.translatedBy.  It wouldn't take a lot of work to change the code, but it would avoid the need to extend the list of exceptions for other languages, give more flexibility, and reduce the code complexity.

r12a: It would also solve the problem for all these other cases:

````
s.author = "By:" // followed by name of author(s)
s.previousAuthors = "Previously by:" // followed by name of previous author(s), before substantive changes were made by the current author
s.modifiedBy = "Changed by:" // person's name appears after colon
````

(there may be some more)

I'd suggest changing the boilerplate text files to say

````
s.translatedBy = "Translator:"+" "
````
For Chinese that would be +"".

I can make this change, if you like. You'd need to remove the changes from this PR.

File: questions/qa-when-lang-neg.zh-hans.html [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+
+<body>
+<header>
+  <nav id="mainNavigation"></nav><script>document.getElementById('mainNavigation').innerHTML = mainNavigation</script>
+  
+  <h1>何时使用语言协商</h1>
+</header>
+
+
+<section>
```
r12a: This section opening tag should be dropped down so that it appears immediately before the h2 just below.

questions/qa-when-lang-neg.zh-hans.html
```diff
@@ -0,0 +1,138 @@
...
+  <div id="audience">
+    <div id="updateInfo"></div><script>document.getElementById('updateInfo').innerHTML = g.updated</script>
+    </div>
+    
+  <h2 id="question"><a href="#question">问题</a></h2>
```
r12a: For all section beginnings, the id needs to go on the section tag, and the a element needs to be removed from the heading.

javascript/doc-structure/article-2022.js
xfq: Fixed. Thanks.

questions/qa-when-lang-neg.zh-hans.html
xfq: Fixed.

questions/qa-when-lang-neg.zh-hans.html
xfq: Fixed.

questions/qa-when-lang-neg.zh-hans.html
```diff
@@ -50,12 +50,12 @@ <h1>何时使用语言协商</h1>
...
 
 
-<section>
   <div id="audience">
     <div id="updateInfo"></div><script>document.getElementById('updateInfo').innerHTML = g.updated</script>
     </div>
     
-  <h2 id="question"><a href="#question">问题</a></h2>
+  <section id="question">
+    <h2><a href="#question">问题</a></h2>
```
r12a: Still need to remove the a tag from the heading.

questions/qa-when-lang-neg.zh-hans.html
```diff
@@ -65,14 +65,14 @@ <h2 id="question"><a href="#question">问题</a></h2>
 </section>
 
 
-<section>
-  <h2 id="answer"><a href="#answer">答案</a></h2>
+<section id="answer">
+  <h2><a href="#answer">答案</a></h2>
```
r12a: Remove link from all the headings.

r12a:
I think we need to split this PR.  Since you're making changes to the javascript structure and boilerplate files, we need to alter the boilerplate text for all languages before merging – otherwise we'll break all the other translations.  I suggest that you remove those changes from this PR, and i'll change the boilerplate files and make those changes (i'll do it now).

That's also cleaner, since it will produce different commits for alterations to the template and for the translation.

r12a:
I implemented the fixes to all the boilerplate-xx.js files and to the xxx-2022.js files, so you can remove those from the pull request now.  Then we might be ready to merge, but i'll double-check after you make the alterations.

questions/qa-when-lang-neg.zh-hans.html
xfq: Sorry, should be fixed now.

questions/qa-when-lang-neg.zh-hans.html
xfq: Fixed.

xfq:
> I implemented the fixes to all the boilerplate-xx.js files and to the xxx-2022.js files, so you can remove those from the pull request now. Then we might be ready to merge, but i'll double-check after you make the alterations.

Removed. Please check.

r12a (approved):

PR #387, opened 2022-06-02
Boilerplate

brandelune:
added untranslated strings.


javascript/boilerplate-text/boilerplate-fr.js
```diff
@@ -55,6 +55,7 @@ s.examplesInAnotherScript = "Ce document contient des exemples dans une autre la
 s.worldMap = "Carte du monde";
 s.searchI18nSite = "Recherche sur le site I18N"; 
 s.translationDisclaimer = 'Ce document est une traduction fournie bénévolement. En cas de divergences ou d\'erreurs, la <a href="'+f.filename+'.en">dernière version originale en anglais</a> fait autorité. Comme indiqué ci-dessous, les <a href="#copyright">droits d\'auteur</a> reviennent au W3C.'
+s.translator = "Traducteur :";
```
r12a, line 58: This line is no longer in the latest version of this file. (s.translatedBy is used instead)

javascript/boilerplate-text/boilerplate-fr.js
```diff
@@ -67,10 +68,10 @@ s.i18nActivityHomePage = "Page d’accueil de l’activité d’internationalisa
 s.home = "Accueil";
 s.contactI18nActivity = "Pour obtenir de l'aide au sujet de l'internationalisation au W3C.";
 s.contact = "Contacter";
-s.aboutI18nActivity = "About W3C Internationalization."
-s.about = "About"
-s.askI18nActivity = "Ask for help or information."
-s.ask = "Ask"
+s.aboutI18nActivity = "Pour en savoir plus sur l'internationalisation au W3C."
+s.about = "En savoir plus"
```
r12a, line 72: Is this the standard translation for "About" in the sense of "About the W3C i18n Activity"?  It's not really a question of learning more, but specifically about information related to the organisation.  Perhaps something like á propos??  (Just asking)

javascript/boilerplate-text/boilerplate-fr.js
```diff
@@ -113,6 +114,7 @@ s.validXHTML = "XHTML 1.0 valide !";
 s.validCSS = "CSS valide !";
 s.codedInUtf8 = "Codé en UTF-8 !";
 
+s.translatedFromEnglishVer = "Traduit d’un contenu en anglais daté du "+dt.enVersion+". Dernière modification de cette traduction le "+dt.thisVersionPlain+" UTC.";
```
r12a, line 117: Not sure why this line is here. The translation already exists at line 126 of your version.

r12a (changes requested):
Thanks for proposing these changes.  See, however, a couple of needed corrections in my comments.

javascript/boilerplate-text/boilerplate-fr.js
```diff
@@ -115,6 +115,15 @@ s.validCSS = "CSS valide !";
 s.codedInUtf8 = "Codé en UTF-8 !";
 
 s.translatedFromEnglishVer = "Traduit d’un contenu en anglais daté du "+dt.enVersion+". Dernière modification de cette traduction le "+dt.thisVersionPlain+" UTC."; 
+// this block of strings should include any whitespace needed after the colon
+// or its equivalent, eg. "By: " in english, but "作者：" in chinese
+s.author = "Par : "; // followed by name of author(s)
+s.previousAuthors = "Précédemment par : "; // followed by name of previous author(s), before substantive changes were made by the current author
+s.modifiedBy = "Modifié par : "; // person's name appears after colon
+s.translatedBy = "Traducteur : ";
```
r12a, line 123: Fwiw, this came up recently during the Galician translation. Since the translator was female, she wondered whether the translation should be Traductor/traductora (here i guess that equates to Traducteur/traductrice).  We finally settled on Traducido por (i guess in French, Traduit par) to keep things simple.  Should that apply here ?

javascript/boilerplate-text/boilerplate-fr.js
brandelune: I see. I updated the data that was in English in the French file without checking the English file... My bad.

javascript/boilerplate-text/boilerplate-fr.js
brandelune: Indeed. There is a new gender-neutral term in French "Traductaire", let me think about it.

brandelune:
I'll update the file with the contents of the English translation. Sorry for the hassle.

javascript/boilerplate-text/boilerplate-fr.js
r12a: Also note that we try to keep these translations as short as possible.  It's fine if there's no one word translation in French, but the shorter we can keep it the better.

PR #389, opened 2022-06-07
Adjust the Chinese font stack

xfq:
Latin fonts are placed before Chinese fonts because some Chinese fonts have ugly English/Cyrillic/Greek glyphs.

For more background, see https://github.com/w3c/clreq/issues/408#issuecomment-1103501409

### DIFF
--- a/style/article-2022.css
+++ b/style/article-2022.css
@@ -23,6 +23,20 @@
     font-style: normal;
     }
 
+/* From https://github.com/w3c/clreq/blob/gh-pages/local.css */
+@font-face {
+	font-family: 'Punctuation SC';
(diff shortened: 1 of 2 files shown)


r12a (changes requested):
I think the @font-face declarations need to appear at the very beginning of the CSS style sheet, otherwise they don't work.

r12a:
ie. before the declaration for the body tag.

xfq:
Done.

r12a (approved):

xfq:
Thank you. Do we need to upload these files to the CVS server?

r12a:
Already done.

PR #392, opened 2022-06-08
Fix some typos on questions/qa-when-lang-neg

### LINKED ISSUE #390: [questions/qa-when-lang-neg]  Possible typos
https://www.w3.org/International/questions/qa-when-lang-neg.en.html

> Suppose Sylvia visits www.example.be and gets Flemish (situation 2 or 3). She then clicks on the German control and reads on, no real trouble. But she then clicks on a link to get to an interesting page within the site. Oops, Flemish again! Fortunately, the German control is still there, but after a couple of such detours she's getting understandably frustrated. Can't www.example.be just remember that she can't read Flemish? What is needed here is some *stickyness* of the explicit language selection.

"stickyness" should probably be "stickiness".

> Note however than some measure of cultural adaptation (e.g. changing the currency) does not necessarily make pages non-equivalent

"Note however than" should probably be "Note however that".

JohnTitor:
Fixes #390

### DIFF
--- a/questions/qa-when-lang-neg.en.html
+++ b/questions/qa-when-lang-neg.en.html
@@ -54,7 +54,7 @@ f.additionalLinks = ''
     <div id="updateInfo"></div><script>document.getElementById('updateInfo').innerHTML = g.updated</script>
     </div>
     
- <section id="question">
+<section id="question">
 <h2>Question</h2>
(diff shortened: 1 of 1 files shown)


r12a:
Thanks!

PR #393, opened 2022-06-12
Optimize the png files

xfq:


r12a:
How did you get these reductions?  (I had already reduced using ImageOptim)  Are they still visually identical?

xfq:
I used [TinyPNG](https://tinypng.com/). I have compared both locally and [online](https://github.com/w3c/i18n-drafts/pull/393/files) and can't see a significant difference.

r12a:
Ok. Let's merge it. I'm a little suspicious (particularly of the arabic text image) because i already optimised them and got nowhere near that kind of compression. So i'll duplicate my local files and do a comparison after the merge.

himorin:
Looking into color palette, nearby and rarely used colors are summarized into one, like three gray colors from `#656565` to `#676767` into `#666666` (which appears at edge of glyph by antialiasing).

r12a:
Thanks @himorin.  I compared the before and after versions, and can't see any difference.

PR #394, opened 2022-06-13
Typo fix

xfq:


r12a:
I can't see the difference here.

xfq:
The meaning of 子母 is "mother and son", and the meaning of 字母 is "letter" (used for explaining the number 18 stands for the number of letters between the first i and the last n in the word internationalization). It's just a typo fix.

r12a:
Ah. The text is so small that i didn't notice the shape difference. Ok by me then.

PR #395, opened 2022-06-13
Editorial fixes

xfq:


r12a:
I can't judge the Chinese, but otherwise this LGTM.

PR #398, opened 2022-06-15
Update Addison's email

### LINKED ISSUE #396: Update Addison's email address
We should update @aphillips's email address at https://www.w3.org/International/i18n-drafts/nav/about#contacts (and other places).

xfq:
Fix #396.

### DIFF
--- a/getting-started/languagedev_participation.html
+++ b/getting-started/languagedev_participation.html
@@ -140,7 +140,7 @@ f.additionalLinks = ''
 <h2 id="contacts"><a href="#contacts">Contacts</a></h2>
 			<ul>
 				<li><a href="mailto:ishida@w3.org">Richard Ishida (ishida @ w3. org)</a>, i18n Lead</li>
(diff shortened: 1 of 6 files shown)


r12a (changes requested):
You missed the link text in the last 3 files. Otherwise, seems ok.

xfq:
I'm trying adding Netlify to this repo.

The Deploy Preview link above returns 404, because I changed multiple files in this PR, and Netlify didn't know which file to preview (and we don't have an index.html at the top level of the repo). If you want to preview an individual HTML file, you can add the path of it after the Deploy Preview link, for example:

https://deploy-preview-398--i18n-drafts.netlify.app/getting-started/languagedev_participation.html#contacts is **not** 404.

If only one file is modified in a PR, the Deploy Preview link won't be a 404 page, like [this one](https://github.com/xfq/i18n-drafts/pull/1).

PR #400, opened 2022-06-17
Unify spelling of TR

mehmetoguzderin:
This document lists the country Türkiye with both Turkey and Türkiye spelling. Still, when both are present, unless there are other constraints that I accidentally skip, it might be better to use Türkiye everywhere since that now has UN recognition to be primary. This PR aims to do that. Thank you!


r12a:
Argh, i missed those. Thanks for catching them @mehmetoguzderin.

mehmetoguzderin:
@r12a Thank you for the understanding and the merge!

r12a:
Reminder to self: Upload changes to W3C site once CVS connection is sorted out.
Change also made to de, es, ru, and uk translations.

PR #403, opened 2022-06-30
Update qa-lang-priorities

xfq:
* http -> https
* OS X -> macOS
* Update the description of Firefox and Safari
* Update the link to Apache httpd


r12a (changes requested):
LGTM, but can you propagate the changes into the Russian translation too?  (There used to be several other translations, but they are now all out of date, so don't need to be changed for these edits.)

xfq:
@r12a Thanks. Done.

r12a (approved):

PR #411, opened 2022-07-29
align datetime value to contents in time

himorin:
Text within time element says `2015年5月22日`, but datetime attribute has `2016-05-22`. This PR fixes these difference.
These are just test cases or pure example for tate-chu-yoko, so there seems no intention to change between attribute and text.

### DIFF
--- a/articles/vertical-text/index-data/digits-pref.html
+++ b/articles/vertical-text/index-data/digits-pref.html
@@ -41,10 +41,10 @@ figure {
 <body>
 <div id="htmlsrc" lang="ja">
 <figure>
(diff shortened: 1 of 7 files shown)


r12a:
@himorin many thanks for spotting this.

Actually, when i looked at that article i noticed that many of the items listing browser support were out of date.  I have therefore begun applying the latest template to the article (which shows live results in the article itself) and updating the test results.  I'll try to integrate these changes you suggest at the same time. I'll let you know when i get further with the conversion.

r12a:
@himorin i incorporated your changes into https://github.com/w3c/i18n-drafts/commit/75bda2227ddb594a697e12bd011ffda124afb4b1 so i'll close this now.

PR #432, opened 2022-11-30
language-tags: update links

meisterluk:
http://r12a.github.io/apps/subtags/
… moved to …
https://r12a.github.io/app-subtags/


r12a:
@meisterluk thank you very much indeed for trawling through all these files (including the translations) to find these links!

PR #433, opened 2022-12-05
Translate qa-text-processing-vs-metadata into zh-hans

xfq:


r12a:
1. the contrastive highlighting of " what **specific language** it is dealing with for a **specific range of text**." and "about **the document/resource as a whole**." appears to be lost in the chinese version.   Is there something we can do to show that emphasis?  (emphasis dots?  bracketing? colour change? font change? etc)
2. we changed the 'follow us' links to include Mastodon, so you'll need to adapt that in the translation.

Otherwise, LGTM, thanks.

r12a:
> we changed the 'follow us' links to include Mastodon, so you'll need to adapt that in the translation

To be more clear, the boilerplate should handle the new structure, and the visible text, but if the title attribute text needs to be adapted for chinese that will need attention. (Maybe not.)

xfq:
Thanks for your review!

> 1. the contrastive highlighting of " what **specific language** it is dealing with for a **specific range of text**." and "about **the document/resource as a whole**." appears to be lost in the chinese version.   Is there something we can do to show that emphasis?  (emphasis dots?  bracketing? colour change? font change? etc)

I've tweaked the colour and font weight to make it clearer.

> 2. we changed the 'follow us' links to include Mastodon, so you'll need to adapt that in the translation.

This is generated by the boilerplate and has nothing to do with the Chinese translation of this article. I'll check the text of the boilerplate (separately).

r12a (approved):

PR #434, opened 2022-12-05
Change RFC 2616 to RFC 9110

xfq:
For `Content-Language`.


aphillips:
Should this point to STD97 instead (to save the need to update RFC numbers from time-to-time)?

xfq:
@aphillips I thought about this, but there isn't a way to link directly to a certain section (`Content-Language`).

(BCP 47 has a similar problem, so I can only link to sections in RFC 4647 and RFC 5646.)

questions/qa-http-and-lang.de.html
```diff
@@ -76,7 +76,7 @@ <h2><code class="kw" translate="no">meta</code>-Element</h2>
 
 <section id="http_summary">
 <h2>HTTP-Header</h2>
-  <p>Wenn Sie eine Website oder -ressource von einem Server anfordern, sendet der Server verschiedene Informationen zu dem Objekt zurück, das Sie anfordern (Metadaten). Er benutzt dabei ein Format, das als <a class="print" href="http://tools.ietf.org/html/rfc2616#section-14.12">HTTP-Header</a> bezeichnet wird. Eines der Dinge, die Sie in solchen Metadaten finden können, ist sprachbezogen. Achten Sie auf die letzte Zeile im folgenden Beispiel, welches die zu diesem Artikel gehörende HTTP-Antwort zeigt.</p>
+  <p>Wenn Sie eine Website oder -ressource von einem Server anfordern, sendet der Server verschiedene Informationen zu dem Objekt zurück, das Sie anfordern (Metadaten). Er benutzt dabei ein Format, das als <a class="print" href="https://httpwg.org/specs/rfc9110.html#field.content-language">HTTP-Header</a> bezeichnet wird. Eines der Dinge, die Sie in solchen Metadaten finden können, ist sprachbezogen. Achten Sie auf die letzte Zeile im folgenden Beispiel, welches die zu diesem Artikel gehörende HTTP-Antwort zeigt.</p>
```
aphillips: Why `httpwg.org` instead of the more official `rfc-editor.org`? I believe the latter is the official site (per John Kleinsin)?

aphillips:
> @aphillips I thought about this, but there isn't a way to link directly to a certain section (`Content-Language`).
> 
> (BCP 47 has a similar problem, so I can only link to sections in RFC 4647 and RFC 5646.)

Good point, although perhaps a shared strategy could work. The <a> link could be to the specific RFC and then be followed by a Respec xref of \[\[STD97\]\]

questions/qa-http-and-lang.de.html
xfq: In my impression, the usability of `httpwg.org` is better. See also https://github.com/w3c/browser-specs/issues/280 and https://github.com/tobie/specref/issues/672

xfq:
> > @aphillips I thought about this, but there isn't a way to link directly to a certain section (`Content-Language`).
> > (BCP 47 has a similar problem, so I can only link to sections in RFC 4647 and RFC 5646.)
> 
> Good point, although perhaps a shared strategy could work. The link could be to the specific RFC and then be followed by a Respec xref of [[STD97]]

This is an article, not a ReSpec document, but I can add a link to STD 97 manually if you think it's useful. (I have no strong preference.)

questions/qa-http-and-lang.de.html
xfq: I changed the links to rfc-editor.org.

xfq:
Merging per https://www.w3.org/2025/07/03-i18n-minutes.html#506e

xfq:
Copied to CVS as well.

PR #438, opened 2022-12-14
Translate further reading

gunnarbittersmann:
Translates “Characters” and “Using Unicode control codes in text”


r12a:
hi @gunnarbittersmann.  @xfq could you look into this for Gunnar, please?

PR #441, opened 2023-01-07
Update the title of articles

xfq:


getting-started/index.html
```diff
@@ -146,7 +146,7 @@ <h2>Taking it further</h2>
 <p>To further develop your knowledge and encounter other topics we recommend the following.</p>
 <ul>
 <li>
-<p> <a class="highlightLink"  href="/International/i18n-drafts/techniques/authoring-html.en">Internationalization techniques: Authoring HTML &amp; CSS</a><br>Resources on the W3C Internationalization Activity site and elsewhere that help you author HTML and CSS for internationalization.</p>
+<p> <a class="highlightLink"  href="/International/i18n-drafts/techniques/authoring-html.en">Internationalization techniques: Authoring Web pages</a><br>Resources on the W3C Internationalization Activity site and elsewhere that help you author HTML and CSS for internationalization.</p>
```
r12a: Authoring Web pages should either be:
Authoring Web Pages, or
Authoring web pages
(probably the latter)

r12a (changes requested):
Just the one nit.

getting-started/index.html
xfq: I copied it from https://www.w3.org/International/techniques/authoring-html . Should we change the capitalization there too?

r12a: Yes, i think so.

xfq: Thanks. Should be fixed now.

PR #445, opened 2023-02-10
Update the name of the typography index

xfq:


pages/language_framework_overview.html
```diff
@@ -153,7 +153,7 @@ <h2>Text layout index</h2>
 
 <figure id="fig_layout_index">
 <img src="language_framework_overview-data/layout_index.png" alt="">
-<figcaption>A small section in the International text layout and typography index.</figcaption>
+<figcaption>A small section in the Language enablement index.</figcaption>
```
aphillips: Either title case it (`Language Enablement`) or lower case it (`language enablement`). The other occurrences are sentence start, hence the initial cap.

aphillips (changes requested):
Minor nit.

pages/language_framework_overview.html
r12a: If you look at the document title https://www.w3.org/TR/typography/ it is downstyled: ie. Language enablement index.  I don't have very strong views on whether or not to retain the downstyling, although if it is link text or highlighted or marked by punctuation in some other way it may be better to keep it.

Note however that if capitalisation _is_ applied, it needs to be applied to all 3 words in the title, not just "Language Enablement" if the title is being quoted (which probably is the case at least most of the time).

articles/typography/justification.en.html
```diff
@@ -264,7 +264,7 @@ <h3>Other Writing Systems</h3>
         <p class="note">Readers who wish to provide such “more specific information”
           are invited (and strongly encouraged) to 
           <a class="interaction" target="_blank" href="https://github.com/w3c/i18n-drafts/issues/new?title=%5Barticles/typography/justification%5D%20specific_title_here&amp;body=%5Bsource%5D%20(http%3A%2F%2Flocalhost%2Fgit%2Fi18n-drafts%2Farticles%2Ftypography%2Fjustification.en.html)%20%5Ben%5D%0A%0Ayour_comment_here">create a github issue</a>
-          so that we can add information or links to this article and the typography index.</p>
+          so that we can add information or links to this article and the language enablement index.</p>
```
xfq: Discussion with @r12a: change 'l' to uppercase and add a link.

articles/typography/justification.en.html
```diff
@@ -264,7 +264,7 @@ <h3>Other Writing Systems</h3>
         <p class="note">Readers who wish to provide such “more specific information”
           are invited (and strongly encouraged) to 
           <a class="interaction" target="_blank" href="https://github.com/w3c/i18n-drafts/issues/new?title=%5Barticles/typography/justification%5D%20specific_title_here&amp;body=%5Bsource%5D%20(http%3A%2F%2Flocalhost%2Fgit%2Fi18n-drafts%2Farticles%2Ftypography%2Fjustification.en.html)%20%5Ben%5D%0A%0Ayour_comment_here">create a github issue</a>
```
aphillips: Our preview is broken so I can't tell... but `http` and `localhost` (for the URL in the parentheses) might be a mistake?

pages/language_framework_overview.html
aphillips: It might be downstyled in the document itself, but it should be treated as a title elsewhere. Otherwise it is super confusing (c.f. this pull request!... ooh, I used a Latin abbreviation... yikes!!) Or we should mark "Language enablement index" with `<cite>`.

In any case, in English the normal behavior for titlecase is to capitalize the first word and all nouns, pronouns, adjectives, verbs and adverbs, but not articles, conjunctions or prepositions.

aphillips (changes requested):
See discussion of titlecasing. Also, check the link.

articles/typography/justification.en.html
xfq: Done.

articles/typography/justification.en.html
xfq: Thanks. Should be fixed now.

pages/language_framework_overview.html
xfq: I have added a link, so I think it's less ambiguous now. Are you OK with it, @aphillips?

See the figure caption in https://deploy-preview-445--i18n-drafts.netlify.app/pages/language_framework_overview.html#fig_layout_index for the preview.

xfq:
Merging per yesterday's i18n WG Teleconference.

PR #446, opened 2023-02-12
QA escapes: FR translation + minor changes to EN

clavoline:
Submitting a French translation of "questions/qa-escapes" and a few changes to the English version.

1) For consistency purposes:

- Replaced `<span class="leading">` and `</span>` with `<strong class="leading">` and `</strong>` in two instances
- Added `<code>` and `</code>` around character references in four instances

2) Added a link to a section on the same page

3) Replaced three obsolete links

4) Modified two links that pointed to different pages but had the same anchor text


r12a:
@clavoline thank you very much for useful edits to the english !   

I haven't looked at the French yet because the netlify link didn't work.  @xfq is it possible to fix that?  (If not, i'll install the text in a local file to check it.)

r12a:
https://deploy-preview-446--i18n-drafts.netlify.app/questions/qa-escapes.fr.html

r12a:
@clavoline i have now looked at the French translation and it looks good, with one caveat: near the beginning you translated the Unicode name for a character, U+00A0 ESPACE INSÉCABLE.  It's normal to leave the uppercase name exactly as it appears in the standard, however, i know that there is a French version of the Unicode Standard which does have translated character names (and by memory i think you have the right name).  Therefore this may not be a problem.  I just wanted to check with you that you were aware.  Just let me know, and if you want to keep it as is then i'll merge and announce the translation.

clavoline:
Thank you Richard, that's a good question.
Is this the French version of the Unicode Standard you were referring to: http://hapax.qc.ca/Tableaux-15.0.htm?
Note: there are two more Unicode character names in the text, which I translated as well ("U+200F MARQUE DROITE-À-GAUCHE" and "U+3123 LETTRE BOPOMOFO EN"). The names match the recommendations from the Hapax website.

I went through every French article I could find on Google with "site:https://www.w3.org/International/questions" and only one article included Unicode character names: https://www.w3.org/International/questions/qa-html-css-normalization.fr.html

In this article, the names have been translated as "U+00E1 Lettre minuscule latine A accent aigu", "U+0061 Lettre minuscule latine A" and "U+0301 Diacritique accent aigu". These names match the recommendations from the Hapax website.

Do you know whether Unicode character names appear in other French translations?

Based on that translation, for consistency purposes, it would make sense to use the French names (but follow French capitalization rules). However, if you'd rather use the English names (and add a French translation or explanation for each of them), I don't mind.

Thank you,
Gwen
------- Original Message -------

r12a:
> In this article, the names have been translated as "U+00E1 Lettre minuscule latine A accent aigu", "U+0061 Lettre minuscule latine A" and "U+0301 Diacritique accent aigu". These names match the recommendations from the Hapax website.

Thanks for pointing that out. I'll change that page. 

I was indeed referring to the Hapax page, which is the work of Patrick Andries.  Patrick only uses lowercase on that page for notes: the actual character names are all in majuscule.  So if that's where you got your translations from then i think we're already good to go, and i should just merge this. I'll wait for your reaction before doing so.

clavoline:
Thank you! To be transparent, I found those translations on Wikipedia, but they match the Hapax recommendations.

I might have one more change to make. I found a Unicode glossary that translated "supplementary characters" as "caractères complémentaires" (rather than "caractères supplémentaires, which I used in the article - in 4-5 instances). Should I use this term instead or are you fine with either?

Thank you,
Gwen
------- Original Message -------

r12a:
'complémentaires' doesn't sound right to me, and Patrick also uses 'caractères supplémentaires' (see http://hapax.qc.ca/pdf/chapitre-15.pdf), so i think you should stick with that.

r12a:
Oops.  That's wrong - hold on...

r12a:
I just checked Patrick's book on Unicode and he does refer to 'plans complémentaires'.  So i stand corrected. :-)

clavoline:
I understand, I didn't think twice about that term because of all the time and energy spent on "surrogate code points" in the same paragraph :)

Do you want me to make the changes?

------- Original Message -------

r12a:
> Do you want me to make the changes?

You're the translator; i rely on your judgement.  ;-)

clavoline:
@r12a Hi, I made the changes we discussed, but since I'm new to Git, I might have skipped a step. Is there anything else I need to do?

r12a:
@clavoline thanks for the prompt.  Looks good to me. I think we're ready to publish.  I'll take it from here, thanks.

PR #448, opened 2023-02-19
Add syntax highlighting for the code block

xfq:


aphillips:
@xfq Can you describe why we want this or what the issue is? I'm reluctant to add much JS to our docs, but not *opposed* to adding it...

xfq:
@aphillips [language-html](https://www.w3.org/International/i18n-activity/guidelines/editing#syntax-highlighting) is used [here](https://github.com/w3c/i18n-drafts/blob/e1cf0c9df4ee7d4bdc7a5baa8c4b1682bcf190a2/questions/qa-changing-encoding.en.html#LL95C23-L95C36), but the relevant [JS & CSS files](https://www.w3.org/International/i18n-activity/guidelines/editing#files) are not included. Perhaps this single line of code doesn't need syntax highlighting and I can update the PR to remove the `language-html` class.

aphillips (approved):

PR #449, opened 2023-02-22
General update of qa-bidi-unicode

r12a:


questions/qa-bidi-unicode-controls.en.html
```diff
@@ -64,20 +64,25 @@ <h1>How to use Unicode controls for bidi text</h1>
...
-  <p class="question">If I'm unable to use markup to correctly order bidirectional text, what can I do?</p>
-  <div class="insideinfonote">
-    <p class="noprint info">Right-to-left text in code samples is represented by UPPERCASE TRANSLATIONS, and left-to-right text by lowercase. </p>
-    <p>If there is a  nearby <img src="../icons/codelink.gif" alt="View code."> image, click on it to see  how the example looks in your browser. From the page that opens you can  view the source code for the example.</p>
-  </div>
-<p>This article assumes that you are familiar with bidirectional text concepts and managing bidirectional text using HTML markup, but that you need to know how to do similar things with Unicode control characters, such as when writing plain text.  If you are not familiar with bidi in HTML, you may find it helpful to first read through the article <a href="/International/articles/inline-bidi-markup/">Inline markup and bidirectional text in HTML</a>.</p>
-</section>
+
+
+<p>This article looks at how content authors can apply direction metadata to bidirectional text when markup is not available.</p>
```
aphillips, line 69: I'm struggling with whether this is metadata here or not? Perhaps...

```suggestion
<p>This article shows how content authors can overcome common directional problems with bidirectional text when markup is not available.</p>
```

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -164,22 +176,13 @@ <h2>Changing the base direction</h2>
...
-    <aside class="sideinfonote"><p class="info">This is one area where markup has an advantage over control codes, since it looks like isolation will be supported for markup before it is for the control codes.</p></aside>
-    </div>
-    <p>FSI is in the same boat as RLI/LRI as far as browser support is concerned. There is a set of tests for these control codes, with <a href="https://www.w3.org/International/tests/repo/results/rli-etc">results for major browsers</a>.</p>
-    <p>The following example shows how these control characters could be used in
+
+<p>When demarcating the boundaries of the change in base direction, you really want to avoid what's inside the boundaries interacting with what's outside – ie. you want to <em>isolate</em> it. For this reason, in an ideal world you would want to follow the recommendation of the Unicode Standard to  use RLI and LRI,  and avoid using RLE and LRE.</p>
```
aphillips: We suddenly talk about "boundaries of change" and I find this kind of a leap--I get it, but maybe not the audience.

Perhaps:

```suggestion
<p>When demarcating the boundaries of an inline run of text, you really want to avoid that text interacting with text that's outside – that is, you want to <em>isolate</em> the inline text from the rest of the paragraph. For this reason, in an ideal world you would want to follow the recommendation of the Unicode Standard to use RLI and LRI, and avoid using RLE and LRE.</p>
```

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -188,44 +191,98 @@ <h2>Changing the base direction</h2>
...
     <p>A similar recommendation is valid when dealing with Unicode control characters.</p>
-    <p>Another thing to bear in mind is that ranges need to be nested appropriately. If you have an embedded LTR range in a RTL context, and that LTR range has some RTL text inside it, it won't produce the right result if your ranges are side by side rather than nested. Note how the direction changes are embedded in the following example, rather than side by side.</p>
-    <figure class="example">
-    <p><code>the title is <span style="color:red;">&amp;#x202B;</span>AN INTRODUCTION TO <span style="color:red;">&amp;#x202A;</span>c++<span style="color:red;">&amp;pdf;</span><span style="color:red;">&amp;pdf;</span> in arabic.</code></p>
-    </figure>
+    <p>However, it's important to bear in mind is that ranges need to be nested appropriately. If you have an embedded LTR range in a RTL context, and that LTR range has some RTL text inside it, it won't produce the right result if your ranges are side by side rather than nested. Note how the direction changes are embedded in the following example, rather than side by side.</p>
```
aphillips: The extra `is` reads funny. Also would suggest changing `side by side` => `side-by-side`

```suggestion
    <p>However, it's important to bear in mind that ranges need to be nested appropriately. If you have an embedded LTR range in a RTL context, and that LTR range has some RTL text inside it, it won't produce the right result if your ranges are side-by-side rather than nested. Note how the direction changes are embedded in the following example, rather than side-by-side.</p>
```

aphillips (changes requested):
Partial review. Thanks for this! Mostly good, just minor nits so far.

questions/qa-bidi-unicode-controls.en.html
klensin: Richard, as a very high level comment, this reads as if it is addressed to someone who is already quite familiar with a large fraction of the issues and for whom you are just filling in a few details.   Depending on your expectations of the audience, that might be ok although I think it should be made more explicit.  The article "[Inline markup and bidirectional text in HTML](https://deploy-preview-449--i18n-drafts.netlify.app/International/articles/inline-bidi-markup/)." might help with that, but the link gets a 404.

The situation is further complicated by some of the examples, which either don't work or leave me thoroughly confused.  Take, e.g., the set that conclude the "Changing the base direction" section.   The proposed fix,
   title="'&#x2067;FREQUENTLY ASKED QUESTIONS&#x2069;' [in persian]"
Says "FREQUENTLY ASKED QUESTIONS" and not "(FAQ)" as shown above, and does not show the string in Persian at all.   Or, if the string "FREQUENTLY ASKED QUESTIONS" is intended to substitute for the Persian, "(FAQ)" is missing.   I can deduce what was probably intended, but I suspect only because I knew what was going on before trying to read the text.

A few paragraphs further up, the reader is told "For this reason, in an ideal world you would want to follow the recommendation of the Unicode Standard to use RLI and...".  If I am to follow those recommendations, I would want to know where I can read that actual text, but no such citation appears to be available.

The first paragraph under "Tightly wrapping..." points me to "[Inline markup and bidirectional text in HTML]: (https://deploy-preview-449--i18n-drafts.netlify.app/International/articles/inline-bidi-markup/)  but that appears to be the same dead link as the one mentioned earlier.  

In addition, unless one has a rather good understanding of what is going on and what interactions might occur, a sentence like "A similar recommendation is valid when dealing with Unicode control characters." is fairly close to meaningless.

Then, under "Dealing with spillover ...", there appears to be an abrupt  and unexplained switch from Arabic to Hebrew and then back to Arabic.

Dare I ask why, if the start and end point pairs are generally preferred to the use of RLM and LRM (the text seems to imply that but does not quite say it), HTML provides named entities for those two but not for RLI, PDI, etc.?

Will the reader know what you mean by a "neutral character"?  And that "neutrals" refers to the same thing?

The tutorial, [Creating HTML Pages in Arabic, Hebrew and Other Right-to-left Scripts](https://deploy-preview-449--i18n-drafts.netlify.app/International/tutorials/bidi-xhtml/) is another dead link as are all of the other links under "Further reading".

The draft is probably in need of another pass.

r12a: > Dare I ask why, if the start and end point pairs are generally preferred to the use of RLM and LRM (the text seems to imply that but does not quite say it), HTML provides named entities for those two but not for RLI, PDI, etc.?

It's not for want of asking, but the HTML5 editors are against it.

questions/qa-bidi-unicode-controls.en.html
r12a: It's not information that the user reads as part of the text (in fact they are invisible anyway), but it provides information about the text.  I think  that qualifies as metadata.  If it were in HTML, we'd be talking about the dir attribute values - for me that's metadata too.

questions/qa-bidi-unicode-controls.en.html
r12a: Changed.

questions/qa-bidi-unicode-controls.en.html
r12a: All points addressed.

aphillips: > > Dare I ask why, if the start and end point pairs are generally preferred to the use of RLM and LRM (the text seems to imply that but does not quite say it), HTML provides named entities for those two but not for RLI, PDI, etc.?
> 
> It's not for want of asking, but the HTML5 editors are against it.

This sounds like something we should revisit with them.

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-<meta charset="utf-8" />
+<meta charset="utf-8">
 <title>How to use Unicode controls for bidi text</title>
-<meta name="description" content="If I'm unable to use markup to correctly order bidirectional text, what can I do?" />
+<meta name="description" content="If I'm unable to use markup to correctly order bidirectional text, what can I do?">
```
aphillips, line 6: Is there a reason to phrase this backwards?

```suggestion
<meta name="description" content="What can I do if I'm unable to use markup to correctly order bidirectional text?">
```

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -164,138 +178,230 @@ <h2>Changing the base direction</h2>
...
-    <figure class="example">
-    <p><code>the title is <span style="color:red;">&amp;#x202B;</span>AN INTRODUCTION TO <span style="color:red;">&amp;#x202A;</span>c++<span style="color:red;">&amp;pdf;</span><span style="color:red;">&amp;pdf;</span> in arabic.</code></p>
-    </figure>
+ 
+<p>In some cases the bidi algorithm copes fine with bidirectional text, and in others it needs some help. In <a href="https://www.w3.org/International/articles/inline-bidi-markup/#tightwrap"><cite>Inline markup and bidirectional text in HTML</cite></a><cite> </cite>we make the case that the easiest approach to marking up bidirectional text is to put markup at the start and end of each directional change in the text. This doesn't do any harm, it avoids the likelihood of missing a situation where markup is needed, and it makes the life of the content author much simpler. You should wrap the relevant text <a class="termref" href="https://www.w3.org/International/articles/inline-bidi-markup/#tightwrap">tightly</a>.</p>
```
himorin:
```suggestion
<p>In some cases the bidi algorithm copes fine with bidirectional text, and in others it needs some help. In <a href="https://www.w3.org/International/articles/inline-bidi-markup/#tightwrap"><cite>Inline markup and bidirectional text in HTML</cite></a> we make the case that the easiest approach to marking up bidirectional text is to put markup at the start and end of each directional change in the text. This doesn't do any harm, it avoids the likelihood of missing a situation where markup is needed, and it makes the life of the content author much simpler. You should wrap the relevant text <a class="termref" href="https://www.w3.org/International/articles/inline-bidi-markup/#tightwrap">tightly</a>.</p>
```

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -92,19 +99,19 @@ <h2>Changing the base direction</h2>
...
         </tr>
-        <tr>
+      <tr>
           <td>LRI</td>
           <td>LEFT-TO-RIGHT ISOLATE</td>
           <td>U+2066</td>
           <td>dir = &quot;ltr&quot;</td>
-          <td>sets base direction to LTR and isolates the embedded content from the surrounding text</td>
-        </tr>
+          <td>ditto, but for LTR</td>
```
xfq: It says "ditto" but it's the first row in the table. Maybe we should swap it with the row below?

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -118,7 +125,7 @@ <h2>Changing the base direction</h2>
           <td>LEFT-TO-RIGHT EMBEDDING</td>
           <td>U+202A</td>
           <td>dir = &quot;ltr&quot;</td>
```
xfq: Why adding spaces around the equal sign? I think if the reader copies the code to other places, it won't work.

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -164,138 +178,230 @@ <h2>Changing the base direction</h2>
...
-        <tr>
-          <td>PDI</td>
-          <td>POP DIRECTIONAL ISOLATE</td>
-          <td>U+2069</td>
-          <td>end tag</td>
-          <td>used for RLI, LRI or FSI</td>
-        </tr>
       </tbody>
     </table>
     <p>These characters are invisible, although in some editors it may be possible to show symbols that represent them. You could also use character escapes to represent them, such as <code>&amp;#x2067;</code>, but in bidirectional source text you may find that the characters in the escape don't stay together. (See <a href="https://www.w3.org/TR/2014/NOTE-i18n-html-tech-bidi-20140603/#bidisource"><cite>Problems with bidirectional source text in markup</cite></a> for more on this.)</p>
```
xfq: I think we have deprecated this document. It might be better not to link to it.

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -164,138 +178,230 @@ <h2>Changing the base direction</h2>
...
+
+<figure class="example">
+<blockquote><code>
+<bdo dir="ltr">title=&quot;'<span style="color: green;" title="RLE">&amp;#x2067;</span>FREQUENTLY ASKED QUESTIONS (faq)<span style="color: red;" title="PDF">&amp;#x2069;</span>' [in persian]&quot;</bdo>
+</code></blockquote>
+</figure>
+
+<div id="out_tooltip" class="browser_output">
+<p class="browser_output_title">Output in your browser (mouse over for result):</p>
+<div class="browser_output_example" lang="ja">
```
xfq: Why is this example (and other examples) tagged `lang="ja"`?

questions/qa-bidi-unicode-controls.en.html
r12a: Thanks for spotting that. Fixed.

questions/qa-bidi-unicode-controls.en.html
r12a: Good point. Fixed.

questions/qa-bidi-unicode-controls.en.html
r12a: Oops. Fixed.

questions/qa-bidi-unicode-controls.en.html
r12a: Mm. I thought we have already copied that section to another document, but i'm not seeing it.  So i'm going to port the information into a new article, and then we can point to that.

r12a: This now links to the future location of the bidi-source article.

questions/qa-bidi-unicode-controls.en.html
```diff
@@ -64,25 +64,32 @@ <h1>How to use Unicode controls for bidi text</h1>
...
+    <p>You will still need to use markup to establish the default direction for a document as a whole (eg. in the <code class="kw" translate="no">html</code> tag), and to change direction for block container elements. Because control codes don't cross paragraph (read as <span class="qterm">block element</span>) boundaries, and because control codes cannot manage inheritance and scoping through the markup hierarchy, they are only appropriate for inline use.</p>
+<p>For example, although a <code class="kw" translate="no">title</code> element in the header of an HTML page can contain no markup, it is still possible to set a default base direction on the <code class="kw" translate="no">title</code> element itself (or for it to inherit the direction set on the <code class="kw" translate="no">html</code> tag). This article is about how to apply directional changes <em>within</em> the <code class="kw" translate="no">title</code> attribute text (or inline in other plain text situations), where needed, or how to apply a direction to attribute text which is different from that of the surrounding element, and so on. It also applies to plain text formats such as WebVTT and CSV.</p>
```
aphillips: Nit: I would change:

> ... although a `title` element in the header of an HTML page can contain no markup...

to read:

> ... although a `title` element in the header of an HTML page cannot contain markup...

(or possibly _additional markup_). The phrase "can contain no" just reads oddly to me.

---

I would also change:

> This article is about how to apply directional changes _within_ the `title` attribute text (or inline in other plain text situations)...

to be the other way around, e.g.:

> This article is about how to apply directional changes _within_ inline (or other plain-text) situations such as found in the `title` element...

Note that you call it the "title attribute" in this sentence but were talking about the title element previously.

r12a: Fixed by https://github.com/w3c/i18n-drafts/pull/449/commits/639da7c9caaa113a9bf3d82b1121338134733efd

PR #451, opened 2023-02-27
Translate qa-changing-encoding into zh-hans

xfq:
The Netlify Deploy Preview is 404 because I modified many files and Netlify didn't know which one to show, if we had an `index.html` at the root of this repository, the Page Not Found error wouldn't appear.

To see the preview, just add the corresponding path after the Netlify URL, i.e., https://deploy-preview-451--i18n-drafts.netlify.app/questions/qa-changing-encoding.zh-hans.html

### DIFF
--- a/questions/qa-changing-encoding-data/translations.js
+++ b/questions/qa-changing-encoding-data/translations.js
@@ -1,9 +1,9 @@
 var trans = { }
 
-trans.versions = ['bg','en','fr', 'it']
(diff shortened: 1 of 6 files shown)


r12a:
I don't see any problems, but i didn't look at the rendered page because netlify preview doesn't work – any chance of fixing that for future reference?

xfq:
Thanks, @r12a. See the (modified) first comment for the preview link.

r12a (approved):

PR #453, opened 2023-03-05
Update the Unicode usage

xfq:
per https://w3techs.com/technologies/history_overview/character_encoding/ms/y


r12a:
LGTM

r12a:
It may be worth also changing the article (https://www.w3.org/International/questions/qa-who-uses-unicode.en.html) at the same time.  I suggest changing:

> By January 2021, that figure had risen to 96.1%.

to 

> By January 2021, that figure had risen to 96.1%, and by Jan 2023 the figure was 97.9%.

PR #458, opened 2023-03-19
Update links to the HTTP specification

xfq:


aphillips (approved):
LGTM

questions/qa-accept-lang-locales.de.html
```diff
@@ -133,7 +133,7 @@ <h2>Literaturhinweise</h2>
 
 <ul id="full-links">
     <li>
-      <p><a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4">Was die HTTP-Spezifikation zu Accept-Language sagt</a> </p>
+      <p><a href="https://www.rfc-editor.org/rfc/rfc9110.html#field.accept-language">Was die HTTP-Spezifikation zu Accept-Language sagt</a> </p>
```
aphillips, line 136: Should we reference STD97 instead?

xfq: The problem is that we can't reference the `Accept-Language` request HTTP header of STD 97. We can only reference the section in RFC 9110.

xfq:
These links appear to have been updated. Closing.

PR #459, opened 2023-03-19
Update the link to the language enablement page

xfq:


articles/css3-text/index.en.html
```diff
@@ -290,7 +290,7 @@ <h2>You can help</h2>
   <p>You can help move the specifications forward by reviewing and commenting on the public drafts that are made available. If you have
       expertise in this area, you might also consider participating in the CSS or Internationalization Working Groups to help move the work along. </p>
     <p>A
-      significant amount of specification work benefitted from the <a href="https://www.w3.org/TR/jlreq/">Japanese Layout Requirements</a> document, and work is going on to <a href="https://www.w3.org/International/layout">document requirements for other scripts</a>. You can contribute to that work, if you have expertise in that area.</p>
+      significant amount of specification work benefitted from the <a href="https://www.w3.org/TR/jlreq/">Japanese Layout Requirements</a> document, and work is going on to <a href="https://www.w3.org/International/i18n-drafts/nav/languagedev">document requirements for other scripts</a>. You can contribute to that work, if you have expertise in that area.</p>
```
aphillips, line 293: The URL seems less good than the old one? Why would the drafts version be pointed to vs. the old location?

aphillips (commented):
Curious why we're making this change...

articles/css3-text/index.en.html
r12a: `i18n-drafts` is the name of the GH repo where these pages reside - they are no longer on CVS.  Adding the repo name was requested when we put together the redirects from www.org, since it helps us know where to find the source text.  This has been the standard location for at least a year now - i guess you missed the memo.  (Think of i18n-drafts like internet drafts.)

xfq:
Merging, per discussion with @r12a offline.

PR #469, opened 2023-04-24
Update qa-html-di

vinhill:
I tested [changing an elements directionality](https://www.w3.org/International/questions/qa-html-dir#userexplicit) with the latest firefox release and the keyboard-shortcut does update the value of the elements `dir` attribute.


r12a:
@vinhill thanks for this correction.  Looking in the inspector, i agree that the direction is now set by the keypress.

However, i'm not seeing Firefox picking up the direction in dirname  so i think it would be worth me adding a note to that effect (as well as removing info about Internet Explorer).  Do you see the same (see https://www.w3.org/International/i18n-tests/results/the-dirname-attribute)?

vinhill:
Yes, dirname is not implemented in Firefox ([Bugzilla entry](https://bugzilla.mozilla.org/show_bug.cgi?id=675943)). Though we are currently working on fixing this.

r12a:
@vinhill It sounds like you're working on fixing dirname in the near term. If so, i guess it's not worth adding a note. Do you expect dirname to be supported in a few months from now?

vinhill:
@r12a I do hope so and can notify you otherwise.

r12a:
@vinhill thanks again for proposing this PR.  In the end, i had other edits to make to that article so i made the changes myself rather than pulling this in.

PR #488, opened 2023-07-04
Update respec_2022.css

r12a:
Add styles for transcriptions, and colour brown for unicode names.


r12a:
Raised as a PR so that the changes are visible, but i'll take any comments after merging, so that the effect is available for the Tibetan doc.

PR #494, opened 2023-07-24
Revise the Chinese translation

xfq:
The previous Chinese translation was not easy to understand, so I tried to revise it.


xfq:

PR #499, opened 2023-07-27
Translate qa-css-lang into zh-hans

xfq:
[Preview](https://deploy-preview-499--i18n-drafts.netlify.app/questions/qa-css-lang.zh-hans.html)

### DIFF
--- a/questions/qa-css-lang-data/translations.js
+++ b/questions/qa-css-lang-data/translations.js
@@ -1,6 +1,6 @@
 var trans = { }
 
-trans.versions = ['de','en'];
+trans.versions = ['de','en','zh-hans'];
 
 trans.outofdatetranslations = [];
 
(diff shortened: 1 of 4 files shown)


r12a:
I think raising the XML section to h2 is good.

I noticed, though, that the "Kaithi, Kai" setting in the earlier examples was not applied in the example in the section Generic class or id selectors.  Unless you know of a reason, perhaps we should make that example consistent with the others?

xfq:
> I noticed, though, that the "Kaithi, Kai" setting in the earlier examples was not applied in the example in the section Generic class or id selectors. Unless you know of a reason, perhaps we should make that example consistent with the others?

Thanks. Fixed.

r12a (approved):

xfq:

PR #506, opened 2023-07-30
Change Thaana to Dhivehi

### LINKED ISSUE #501: [tutorials/bidi-xhtml/index]  Language name vs script name?
[source] (https://www.w3.org/International/tutorials/bidi-xhtml/) [en]

> Right-to-left scripts are used by numerous languages, including Arabic, Hebrew, Pashto, Persian, Sindhi, Syriac, Thaana, Urdu, Yiddish, etc

Most of the names listed here are language names (some are also script names, such as Arabic, Hebrew, and Syriac), but there is one exception, Thaana, which can only represent the script. We should probably change it to Dhivehi (Maldivian).

xfq:
Fix #501.


r12a (approved):

xfq:
Merging. Also copied to CVS.

PR #510, opened 2023-08-03
Revise the definition of bidirectional text

xfq:
- Copy it from https://www.w3.org/TR/i18n-glossary/#dfn-bidirectional-text because it is easier to understand than the one in the tutorial.
- Remove the RTL scripts and languages, since the Objectives section on the right already has it.


xfq:
If/When this is approved, I'll also update the definitions in the translations.

r12a (approved):

PR #520, opened 2023-08-24
Update the "backwards deletion" Q+A article

aphillips:
This is w3c/i18n-actions#6.

Preview it here: https://aphillips.github.io/i18n-drafts/questions/qa-backwards-deletion.en.html


xfq:
I updated the article to match the latest template.

We also need to:

* add a `qa-backwards-deletion/translations.js` file
* add `alt` attribute to the images

questions/qa-backwards-deletion.en.html
```diff
@@ -30,24 +30,82 @@
...
+	    margin-inline: 0.25em;
+        vertical-align: bottom;
+	}
+
+	.uname {
+	    font-family: Arial, monospace;
+		font-size: 0.75em;                    
+		letter-spacing: 0.03em;
+		color: brown;
+	}
```
xfq: I think this should be in the [shared CSS file](https://github.com/w3c/i18n-drafts/blob/gh-pages/style/article-2022.css), instead of this article.

questions/qa-backwards-deletion.en.html
```diff
@@ -86,15 +144,15 @@ <h2 id="terminology">Terminology</h2>
...
   <p><img src="./qa-backwards-deletion-data/text-selection.png"></p>  
   
-  <p><dfn>Text selection</dfn> is the progressive process of selecting at the character level. Text selection on modern systems is done contiguously in memory, using the logical order of characters. When working with bidirectional text, this can produce multiple regions of highlighted text on the screen, although the characters are always adjacent in memory. (Some user interfaces also provide word-by-word or sentence-by-sentence selection, which we won't concern ourselves with here.)</p>
+  <p><dfn>Text selection</dfn> is the progressive process of selecting text in a document. Text selection on modern systems is done contiguously in memory, using the logical order of characters. When working with bidirectional text, this can produce multiple regions of highlighted text on the screen, although the characters are always adjacent in memory. (Some user interfaces also provide word-by-word or sentence-by-sentence selection, which we won't concern ourselves with here.)</p>
   </section>
   <section>
   <h3 id="backwards_deletion">Backspacing ("backwards deletion")</h3>
```
xfq, line 151: It only mentions RTL and LTR text, what about vertical text?

questions/qa-backwards-deletion.en.html
```diff
@@ -185,17 +245,34 @@ <h3 id="combining_marks">Combining marks</h3>
...
   
-  <p>Tamil presents the same concept in a visually more striking way. The syllable 'ko' consists of a two character sequence (U+0B95 U+0BCB), which looks as if it is made of three units. What's more the base character is the one in the middle. These characters still behave the same as those in Hindi (or other languages): cursoring, selection, and forward deletion move over the pair as a single unit. Backspacing deletes the combining mark first:</p>
+  <p>Tamil presents the same concept in a visually more striking way. The syllable <samp>&#xb95;&#xbcb;</samp> (pronounced like 'ko') looks as if it is made of three units. However, it consists of a two character sequence (U+0B95 U+0BCB). What's more the base character is the one visually in the middle. These characters still behave the same as those in Hindi (or other languages): cursoring, selection, and forward deletion move over the pair as a single unit. Backspacing deletes the combining mark first.</p>
```
xfq:
```suggestion
  <p>Tamil presents the same concept in a visually more striking way. The syllable <span lang="ta">&#xb95;&#xbcb;</span> (pronounced like 'ko') looks as if it is made of three units. However, it consists of a two-character sequence (U+0B95 U+0BCB). What's more the base character is the one visually in the middle. These characters still behave the same as those in Hindi (or other languages): cursoring, selection, and forward deletion move over the pair as a single unit. Backspacing deletes the combining mark first.</p>
```

The [samp](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-samp-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) sample or quoted output from another program or computing system. We should probably just use a `span` element with `lang="ta"`.

questions/qa-backwards-deletion.en.html
```diff
@@ -185,17 +245,34 @@ <h3 id="combining_marks">Combining marks</h3>
...
+  
+    <div class="try">
+	  <h4>Try it in your browser</h4>
+	  <p><input id="tryTamil" type="text" name="tryTamil" lang="ta" class="try" value="&#xb95;&#xbcb;"></input>
+	  <button type="button" onclick="reset('tryTamil', '&#xb95;&#xbcb;')" class="reset">Reset</button></p>
+  </div>
   
-  <p>INSERT TAMIL HERE</p>
+  <p>Indic scripts, such as the Devanagari and Tamil examples above, are not the only scripts affected by this. The same can be found for combining marks in many languages. For example, the first cluster in this Thai word <q lang="th">ห้องน้ำ</q> has similar behavior. The end of this word shows additional complexity: the <span class="codepoint" translate="no"><bdi lang="th">&#xe33;</bdi><code class="uname">U+0E33 THAI CHARACTER SARA AM</code></span> appears as a separate typographical unit for effects such as inter-character spacing, but correctly behaves as a single grapheme for the purposes of selection, cursoring, and forward deletion.</p>
```
xfq: Thai is also an Indic script [according to Wikipedia](https://en.wikipedia.org/wiki/Brahmic_scripts#Comparison). It seems that as long as a script is descended from the Brahmi script of ancient India, it can be called an Indic script.

questions/qa-backwards-deletion.en.html
```diff
@@ -218,22 +295,86 @@ <h3 id="combining_marks">Combining marks</h3>
...
+  <p>Another counter case appears in some Indic script langauges where some conjuncts are formed with multiple base characters. An example from the Devanagari script is the syllable <em>kshi</em> (&#x0915;&#x094d;&#x0937;&#x093f;) which is formed using the sequence U+0915 U+094D U+0937 U+093F. The characters U+0915 and U+0937 are both base characters and technically this forms two grapheme clusters. However, in many fonts (and to many users) this character sequence forms a single "shape" perceived to be a single unit of text. In spite of this perception, though, on some browsers the user can both cursor into the conjunct and forward delete only a part of the sequence.</p>
```
xfq:
```suggestion
  <p>Another counter case appears in some Indic script languages where some conjuncts are formed with multiple base characters. An example from the Devanagari script is the syllable <em>kshi</em> (&#x0915;&#x094d;&#x0937;&#x093f;) which is formed using the sequence U+0915 U+094D U+0937 U+093F. The characters U+0915 and U+0937 are both base characters and technically this forms two grapheme clusters. However, in many fonts (and to many users) this character sequence forms a single "shape" perceived to be a single unit of text. In spite of this perception, though, on some browsers the user can both cursor into the conjunct and forward delete only a part of the sequence.</p>
```

questions/qa-backwards-deletion.en.html
```diff
@@ -185,17 +245,34 @@ <h3 id="combining_marks">Combining marks</h3>
   
   <p><img src="./qa-backwards-deletion-data/backwards-deletion.png" alt="Hindi backwards deletion"></p>
   
+  <div class="try">
+	  <h4>Try it in your browser</h4>
+	  <p>Try selecting, cursoring, deleting, and backspacing with this word in Hindi (in the Devanagari script). The word means "Unicode" and contains <em>four</em> graphemes and <em>seven</em> Unicode code points.</p>
+	  <p><input id="tryHindi" type="text" name="tryHindi" lang="hi" class="try" value="&#x92f;&#x942;&#x928;&#x93f;&#x915;&#x94b;&#x921;"></input>
```
xfq, line 244: Note that `input` elements don't need a close tag. (Same for other instances.)

questions/qa-backwards-deletion.en.html
aphillips: That's fairly hair-splitting, but I changed the text anyway.

questions/qa-backwards-deletion.en.html
aphillips: done

questions/qa-backwards-deletion.en.html
aphillips: Added a note about vertical text

questions/qa-backwards-deletion.en.html
aphillips: Done

questions/qa-backwards-deletion.en.html
aphillips: Actually... I added a section about vertical text including an example.

questions/qa-backwards-deletion.en.html
```diff
@@ -218,22 +282,66 @@ <h3 id="combining_marks">Combining marks</h3>
...
+	  
+	  <p>This field contains the Devanagari conjunct <em>kshi</em></p> 
+	  <p><input id="tryConjunct" type="text" name="tryConjunct" lang="hi" style="font-size:24pt;" value="&#x0915;&#x094d;&#x0937;&#x093f;"></input>
+	  <button type="button" onclick="reset()" class="reset">Reset</button></p>
+  </div>  
+  
+</section>
+
+<section>
+	<h3>Vertical text</h3>
```
xfq: We should add an id for this section.

questions/qa-backwards-deletion.en.html
```diff
@@ -358,36 +349,16 @@ <h4>Try it in your browser</h4>
...
-<input type=submit>
-
-
-</form>
+	<h3>Vertical text</h3>
+	
+      <p>Vertical text works in a similar way to horizontal text with regard to text selection, cursoring, deletion, and backspacing. As you might suspect, cursor keys on a keyboard in general move the cursor in the direction of the arrow on the key. This usually means that, in vertical text, "up" is generally preceding the current cursor position logically, while "down" is generally following, while "left" and "right" navigate between lines of text.</p>
+  
+      <!-- Chinese text is from the zh-Hans "What is Unicode?" page -->
+	  <h4>Try it in your browser</h4>
```
xfq, line 346: Since [this is not supported by all browsers yet](https://github.com/w3c/clreq/issues/247), we should probably add a note to explain that this might not be rendered correctly in Chrome and Safari.

style/article-2022.css
```diff
@@ -895,7 +895,25 @@ small {
...
+        padding-inline: 0.25em;
+    }
+	
+    .codepoint img {
+        height: 2em;
+        margin-inline: 0.25em;
+        vertical-align: bottom;
+    }
 
+    .uname {
```
xfq, line 911: If we add `.uname` here, we should probably remove the `.uname` below: https://github.com/w3c/i18n-drafts/blob/d66cd317fd88a414a579c5b6fe3c8ef2046ec2d3/style/article-2022.css#L993-L996

questions/qa-backwards-deletion.en.html
aphillips: Oh, yes, I forgot about that. Fixed.

questions/qa-backwards-deletion.en.html
aphillips: Good catch. Also added one to the exceptions section and cleaned up by adding self-links to `h2` elements (to match the document style--not sure why Richard does this)

style/article-2022.css
aphillips: Done.

questions/qa-backwards-deletion.en.html
r12a: > adding self-links to h2 elements (to match the document style--not sure why Richard does this)

I'm not aware that i do, unless you've been looking at some very old code that i missed converting – or perhaps you're looking at the rendered source code of an article??  The self links are created by the javascript that also creates the TOC.  All that is needed is the id **on the section tag**.

questions/qa-backwards-deletion.en.html
```diff
@@ -78,59 +123,61 @@ <h2 id="intro"><a href="#intro">Introduction</a></h2>
...
   
   <p>Text selection begins much like cursoring, by positioning the cursor at the start (or end) of the desired text and then selecting to the other end of the desired text. This can be done using a pointing instrument, such as a mouse, or using keyboard gestures such as holding "shift" and cursoring through the text. Unlike cursoring, text selection is constrained by the need to select logical characters, so a different number of keystrokes or gestures may be required compared to simple cursoring. This is particularly true for bidirectional text.</p>
   
   <p>Selection using a pointing device, such as a mouse, is subtly different in most implementations than using the cursor keys to extend a selection. When using a pointing device, the selection is entirely logical, between the start and end point of the selection. At least on most physical keyboards, the user can access text selection, usually by holding down the "shift" key while cursoring in the text. As noted before, the cursor keys always move visually and in the indicated direction of the key. For certain bidirectional texts this can mean that the entire text cannot be selected via the cursor keys alone!</p>
```
r12a, line 184: I believe this is incorrect.  I just checked, using some bidi text in my Arabic picker, and when the shift key is held down and the cursor keys used to extend the highlight, the left and right cursor keys extend the highlight in the opposite direction to that observed when simply moving the cursor.

r12a: Try it on Firefox.  Chrome does things differently.

r12a:
> Generally speaking, most text navigation and editing follows the user-perceived character boundaries. For most implementations this corresponds to Unicode's definition of "default extended grapheme cluster" boundaries [[UAX29](https://unicode.org/reports/tr29/)]. The main exception to this is backspacing, which usually follows Unicode code point boundaries in the underlying encoded text (although there are exceptions to this). For the simplest scripts and languages, these often amount to the same thing.

This and other parts of the document strike me as over-simplified and in places incorrect, but there are terminological problems (which we are already familiar with) that cloud the issue.  My experience in working with these things has lead me to view the world in terms of code points, which are grouped into grapheme clusters, which are in turn grouped into orthographic syllables. (I'm in the process of writing that up more clearly, elsewhere...)

I'm inclined to agree with Norbert that this idea of user-perceived character boundaries is too vague and not clearly substantiated enough to be used as the name of a unit of segmentation.  Rather it's merely a way of helping people imagine why code point units are not sufficient _in some cases_. The distinction between grapheme clusters and orthographic syllables is not informed by it's used, but is crucial in the information provided by this article.

My experience has shown that browsers use these 3 different units for text operations such as cursor movement and deletion, depending on the language, and sometimes inconsistently within a single language, but also from browser to browser.  I've been investigating this and writing up results for the various browsers in my orthography notes, under the section "Graphemes".  It may be worth going to https://r12a.github.io/scripts/switch.html and selecting the 'graphemes' segment id, then cycling throught the orthographies using the control "Select an orthography".  You should especially look for the subheading "Browser behaviour", where it exists, to find the results per browser.

(I was wondering whether it would be useful to list behaviour against orthography in a table of some sort – not necessarily in this article, but somewhere.)

That said, it's not clear to me what is your source of authoritative information about how cursoring and deletion should work.  I don't think that it is made clear in the UAX how things should work, but is rather left up to the application to decide the exact mechanism.(?)   Or are you meaning to describe what browsers currently do?  I think it would be good to make that much clearer.

I also think that the article should make it much clearer (actually, i think it's hardly mentioned at all other than for one Thai example) that very different segmentation rules may apply for other operations on the text, such as line breaking, justification, text spacing, and the like – and that this is not an issue, but is useful.

The exceptions section alludes to the importance of orthographic syllables, but this isn't really an exception - even in terms of current browser support.  Again it varies by browsers and by orthography, but it's something that needs to be mentioned either together with or given equal importance to the section entitled "Combining characters".

questions/qa-backwards-deletion.en.html
```diff
@@ -185,17 +232,34 @@ <h3 id="combining_marks">Combining marks</h3>
...
+	  <button type="button" onclick="reset('tryTamil', '&#xb95;&#xbcb;')" class="reset">Reset</button></p>
+  </div>
   
-  <p>Indic scripts, such as the Devanagari and Tamil examples above, are not the only scripts affected by this. The same can be found for combining marks in many languages. For example, the first cluster in this Thai word: <q>คืออะไร</q>. [get better example; demonstrate middle cursor deletion effects in Thai]</p>
+  <p>South-Asian scripts, such as the Devanagari and Tamil examples above, are not the only ones affected by this; similar behavior can be found in any script that employs combining marks. For example, the first cluster in this Thai word <q lang="th">ห้องน้ำ</q> has similar behavior. The end of this word shows additional complexity: the <span class="codepoint" translate="no"><bdi lang="th">&#xe33;</bdi><code class="uname">U+0E33 THAI CHARACTER SARA AM</code></span> appears as a separate typographical unit for effects such as inter-character spacing, but behaves as a single grapheme for the purposes of selection, cursoring, and forward deletion.</p>
```
r12a, line 260: The description of the SARA AM is at the end of the word is not quite correct. It doesn't appear as a single typographic unit for text spacing: rather it is split into a combining mark and a letter, and the space is introduced before the latter – so actually it is split into 2 typographic units, the first of which includes the preceding letter and its combining tone mark.

aphillips: Thanks Richard for your comments.

I'm not unaware that this article doesn't really do the job and I agree with the points in your long comment above. Part of me wants to rubbish the whole thing. A proper job would require a thorough rewrite, which is more work than I think I'm willing to do, so I may look for a volunteer to take it over.

xfq:
I think it might be useful to add an example of IVS. For example, the characters on [this page](https://xfq.github.io/testing/ivs/ivs.xml) are made of two code points (U+9F8D + an ideographic variation selector), but for users, they should be input, selected, and deleted as a whole. Regarding input methods, many input methods can already input IVS. We can mention cursor movement, selection, and deletion here.

aphillips:
The working group elected not to complete work on this QA document. However, I don't want to lose the invested effort. I'm merging the changes for now. I should probably add a visible deprecation too.

PR #522, opened 2023-08-29
French t9n Language Tags in HTML and XML

clavoline:


r12a:
This is very helpful, thanks!  I will merge this as a first step on which to base some additional changes arising from suggestions at https://github.com/w3c/i18n-translations/issues/59

PR #525, opened 2023-09-02
Translate qa-ltr-scripts-in-rtl into zh-hans 

xfq:
[Preview](https://deploy-preview-525--i18n-drafts.netlify.app/questions/qa-ltr-scripts-in-rtl.zh-hans.html)


questions/qa-ltr-scripts-in-rtl.zh-hans.html
```diff
@@ -0,0 +1,425 @@
...
+</div>
+<p>虽然你可能很少需要这样做，但如果你想的话，我们将在这里介绍如何让文本看起来像原始文本。</p>
+
+
+
+<p>就中文和日文而言，你会遇到的大多数现代文本片段都很短，并且通常设计为古风，例如为了让游客回想古代中国和日本。</p>
+
+<p>对于日语，sci.lang.japan新闻组上的一篇<a href="https://www.sljfaq.org/afaq/right-to-left.html">有趣的文章</a>解释说，在明治时期之前，除非有空间限制，文字通常是直排的。19世纪末，横排开始作为直排的严肃替代品，引发了对行内文字书写方向的初步实验。</p>
+<blockquote>
+<p>明治时代（1868-1912），在横书流行之初，出现了一种短暂的书写方式，称为“右横书”，与现代的书写方式——左横书相反。这和阿拉伯语、希伯来语等语言的由右至左横向书写风格类似，在页面左侧换行……这种形式从未被广泛使用，也没有保留下来。</p></blockquote>
```
r12a: In this citation you put quotes around “右横书” but not its lefthand equivalent.  It also seem like you translated into Chinese (which is ok), but removed the original Japanese spelling, which i think we shouldn't do, since it helps those who do understand Japanese to understand the terminology.

questions/qa-ltr-scripts-in-rtl.zh-hans.html
```diff
@@ -0,0 +1,425 @@
...
+</figure>
+</div>
+
+
+
+<section id="use_case_types">
+<h3>两种不同类型的用例</h3>
+
+<p>上面的用例可以分为两类，而我们下面提出的两种解决方案可以满足这两类用例。</p>
+<p><strong>第一种用例，可能也是最有可能出现的一种，是表示说明性文本的简短示例。</strong>在这种情况下，如果字符不以相反的顺序显示，则这个例子将失去其意义。例如，如果你得到了错误的文字排列顺序：“文本<span lang="ber">ⴰⵙⵉⴹⵢⴰⵙ⵰</span>最初是这样写的：<bdo lang="ber" dir="ltr">ⴰⵙⵉⴹⵢⴰⵙ⵰</bdo>”，读者会摸不着头脑。因此，在这种情况下，我们希望确保读者看到的<strong>始终是反向排列的文字</strong>，而我们<strong>通过标记语言</strong>来做到这一点。</p>
```
r12a, line 115: The strong text in these two paras doesn't stand out much.  Do you have any idea about what we should add to the chinese-specific styling to create a more visible distinction?

r12a (changes requested):

questions/qa-ltr-scripts-in-rtl.zh-hans.html
xfq: I kept the original Japanese spelling at first, but I found that this would hinder readers' reading. I think if we want to keep it, we should either display the original Japanese spelling of this word in the form of a tooltip, or add the original English text of this paragraph somewhere (such as below this paragraph or at the end of the article, with a link to it).

questions/qa-ltr-scripts-in-rtl.zh-hans.html
xfq: I think the following style will make it stand out:

```css
font-weight: 900;
color: black;
```

And it should be in our global CSS for `zh`, because other Chinese articles and pages also have this issue.

questions/qa-ltr-scripts-in-rtl.zh-hans.html
xfq: Fixed.

questions/qa-ltr-scripts-in-rtl.zh-hans.html
xfq: Thanks. Fixed now.

r12a (approved):

xfq:

PR #532, opened 2023-09-07
Update the information about isolation

### LINKED ISSUE #518: [questions/qa-bidi-controls]  CSS shim link broken
[source] (https://www.w3.org/International/questions/qa-bidi-controls#reasons) [en]

> Currently, the adoption of those new control codes is lagging in browsers, whereas isolation can be applied already using markup. (At the moment you can use a [CSS shim](https://www.w3.org/International/articles/inline-bidi-markup/index#cssshim), or dedicated elements such as bdi, but browsers are also working on applying isolation to the use of the dir attribute by default.)

The CSS shim link seems broken.

xfq:
Fix #518.


xfq:
Copied to CVS.

PR #533, opened 2023-09-07
Fix typo in CSS

xfq:
Mongolian uses `vertical-lr`.


xfq:

PR #539, opened 2023-09-18
Use browser DevTools to check HTTP headers

### LINKED ISSUE #491: [questions/qa-headers-charset]  Use browser DevTools to check HTTP headers
[source] (https://www.w3.org/International/questions/qa-headers-charset.en) [en]

In addition to the methods listed in the article, another method to check HTTP headers is to use browser DevTools. See:

* https://developer.chrome.com/docs/devtools/network/#search
* https://developer.chrome.com/docs/devtools/network/reference/#headers
* https://firefox-source-docs.mozilla.org/devtools-user/network_monitor/request_details/index.html
* https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/network/reference

Also, the Delorie HTTP Header Viewer mentioned in the article doesn't seem to support https.

xfq:
Fix #491.

I'll update the translations if/when it's approved.

### DIFF
--- a/questions/qa-headers-charset.en.html
+++ b/questions/qa-headers-charset.en.html
@@ -15,12 +15,12 @@ f.previousauthors = '' // as above
 f.modifiers = 'Richard Ishida, W3C' // people making substantive changes, and their affiliation
 f.searchString = 'qa-headers-charset' // blog search string - usually the filename without extensions
 f.firstPubDate = '2003-06-16' // date of the first publication of the document (after review)
-f.lastSubstUpdate = { date:'2023-10-13', time:'07:29'}  // date and time of latest substantive changes to this document
+f.lastSubstUpdate = { date:'2023-10-17', time:'07:29'}  // date and time of latest substantive changes to this document
 f.status = 'published'  // should be one of draft, review, published, notreviewed or obsolete
 f.path = '../' // what you need to prepend to a URL to get to the /International directory
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2023-10-13', time:'10:16'} // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2023-10-17', time:'10:16'} // date and time of latest edits to this document/translation
 f.contributors = '' // people providing useful contributions or feedback during review or at other times
 // also make sure that the lang attribute on the html tag is correct!
 f.sources = '' // describes sources of information
@@ -54,10 +54,8 @@ f.additionalLinks = ''
   <section>
     <h2 class="notoc">Related links</h2>
     <p><a class="print" href="https://validator.w3.org/i18n-checker/">i18n Checker</a></p>
-    <p><a class="print" href="http://webtools.mozilla.org/web-sniffer/">Mozilla Web-sniffer</a></p>
-    <p><a class="print" href="http://www.delorie.com/web/headers.html">Delorie HTTP Header Viewer</a></p>
-    <p><a class="print" href="http://www.rexswain.com/httpview.html">Rex Swain&#x27;s HTTP Viewer</a></p>
-    <p><a class="print" href="http://validator.w3.org/#validate_by_uri+with_options">W3C Markup Validator [extended interface]</a></p>
+    <p><a class="print" href="https://www.rexswain.com/httpview.html">Rex Swain's HTTP Viewer</a></p>
+    <p><a class="print" href="https://tools.keycdn.com/curl">HTTP Header Checker by KeyCDN</a></p>
   </section>
 </section>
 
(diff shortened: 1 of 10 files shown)


questions/qa-headers-charset.en.html
```diff
@@ -80,37 +80,63 @@ <h2>The i18n Checker</h2>
...
     </ul>
     <p>Note: W3C has no relationship to any of these services.</p>
     <p>In the HTTP headers, look for the <code class="kw" translate="no">Content-Type</code> header, and in particular for the <code class="kw" translate="no">charset</code> parameter, e.g.</p>
     <blockquote><code>Content-Type: text/html; charset=utf-8</code></blockquote>
     <p>Note: The <code class="kw" translate="no">charset</code> parameter may not be present. This is okay if your document itself indicates its character
       encoding.</p>
   </section>
-  
+
+  <section id="devtools">
```
r12a, line 101: We already discussed a better heading.

Btw, these days i use some conventions when formatting source code that help with maintenance.  If you'd like to do the same (up to you) i suggest:
1. leaving 5-6 empty lines before a section tag (makes it much easier to skip to or find section starts)
2. leaving one blank line after a heading
3. removing all line indentation, but adding a blank line before each of the following tags: p, ol, li (after the first), etc.

questions/qa-headers-charset.en.html
```diff
@@ -80,37 +80,63 @@ <h2>The i18n Checker</h2>
...
+        <p>Modern desktop browsers usually contain a set of web developer tools, and they can be used to view HTTP headers of a web document and its related resources. Here's how to do it in major browsers.</p>
+
+    <section id="devtools-1">
+      <h3>Chrome, Edge, Firefox, and Opera</h3>
+        <ol>
+          <li>Open DevTools by pressing <kbd>Control+Shift+I</kbd> or <kbd>Command+Option+I</kbd> (Mac).</li>
+          <li>Click the <strong>Network</strong> tab. Right now the Network panel is empty.</li>
+          <li>Reload the page to log some network activity. Each row represents a resource. The top resource is usually the main HTML document.</li>
+          <li>Click a resource and open the <strong>Headers</strong> tab to view the HTTP headers.</li>
+        </ol>
```
r12a, line 113: Since this page is actually about finding the character encoding information, i'd suggest to add another list item to tell the reader to look for the `content-type` value. Otherwise, some readers may get confused by `content-encoding` or `accept-encoding` headers.

questions/qa-headers-charset.en.html
```diff
@@ -80,37 +80,63 @@ <h2>The i18n Checker</h2>
...
+
+    <section id="devtools-safari">
+      <h3>Safari</h3>
+        <ol>
+          <li>First ensure that the <strong>Develop</strong> menu is enabled: choose Safari > Settings, click Advanced, then select “Show Develop menu in menu bar”.</li>
+          <li>Show page resoueces by clicking <strong>Develop</strong> > <strong>Show Page Resources</strong>.</li>
+          <li>Press <kbd>Command+Option+0</kbd> to show the details sidebar.</li>
+          <li>Reload the page to log some network activity.</li>
+          <li>Click a resource from the left sidebar and the HTTP headers will appear in the right sidebar.</li>
+        </ol>
```
r12a, line 126: Suggest same extra list item here.

questions/qa-headers-charset.en.html
```diff
@@ -80,37 +80,63 @@ <h2>The i18n Checker</h2>
...
+    </section>
+  </section>
+
   <section id="validator">
 <h2>Use the W3C Markup Validation Service</h2>
```
r12a: Should we keep this section?  It seems that even the extended interface doesn't show the character encoding information any more.

r12a (changes requested):

questions/qa-headers-charset.en.html
xfq: Thanks for the suggestion. I usually use one blank line after a section tag, and I think that's enough for me.

(Some specs use [auto-generated section names in the comment](https://github.com/w3c/csswg-drafts/blob/f1b985bd66bdc0b73ad6e76468342014e76cfaf8/css-values-5/Overview.bs#L50-L58), but I think that's too much.)

questions/qa-headers-charset.en.html
xfq: Done. Thanks.

questions/qa-headers-charset.en.html
xfq: Fixed.

questions/qa-headers-charset.en.html
xfq: Good catch. I have removed this section.

r12a (approved):

xfq:

PR #543, opened 2023-09-26
 [qa-navigation-select] Fix Chinese translation

xfq:
* The word `select` was translated in the original translation, but it should not be translated
* Add style for Chinese emphasis
* Update links to CLDR
* Add missing `mark` tags
* Other editorial fixes


xfq:

PR #550, opened 2023-10-11
Update the titles of the techniques page

xfq:


r12a:
LGTM as long as it's always clear from the link or the styling that this is the title of a document.

article-list/articlelist.en.html
```diff
@@ -51,7 +51,7 @@
 <div class="content">  
 <h1>Articles &amp; tutorials</h1>
 
-<p>You can also find resources using the page <a href="https://www.w3.org/International/techniques/authoring-html" class="external text" title="https://www.w3.org/International/techniques/authoring-html">Internationalization techniques: Authoring HTML &amp; CSS</a>, which provides more fine-grained access to information.</p>
+<p>You can also find resources using the page <a href="https://www.w3.org/International/techniques/authoring-html" class="external text" title="https://www.w3.org/International/techniques/authoring-html">Internationalization techniques: Authoring web pages</a>, which provides more fine-grained access to information.</p>
```
aphillips: I know that Richard has slightly different preferences, but I think we should discuss them (and add the results to our style guide for posterity).

I tend to think that we should use "Elements of Style" title case, which capitalizes most words (except articles and prepositions). That would make this document:

> Internationalization Techniques: Authoring Web Pages

Separately I would note that W3C tends to capitalize Web, which makes "Authoring Web pages" look odd.

aphillips (changes requested):
See comment about titlecase

article-list/articlelist.en.html
r12a: In fact, i think it was this year, or maybe last, W3C Comm team issued guidance that said that the word 'web' would mostly be lowercase, and only uppercase when referring to '_the_ Web'.  Terms like 'web page' would be lowercase.

@aphillips i had similar thoughts to you, but i'm not sure whether you are advocating that the headings themselves should be capitalised, or just the link text.  I would like to maintain our downstyle approach for the headings themselves, but am inclined to support having capitalised link text.

xfq: I'm OK with either one, as long as it's consistent.

See also:

* https://blog.ap.org/products-and-services/ready-to-lowercase-internet-and-web
* https://w3c.studio24.net/docs/proposed-site-map-v2/#sentence-case-for-navigation-labels-and-headings

xfq:
Per 2024-05-16 teleconference discussion, merging this PR as is.

PR #551, opened 2023-10-12
Fix selflinks by adding section and id

xfq:
I'll update the translations when this is approved.


r12a (approved):

xfq:

PR #552, opened 2023-10-12
`:lang` is not a property

xfq:
It's a pseudo-class.


r12a (approved):

xfq:

PR #557, opened 2023-10-28
Language tags French t9n update

clavoline:
@r12a Submitting the changes again because I think I got it wrong the first time, sorry!

### DIFF
--- a/articles/language-tags/index.fr.html
+++ b/articles/language-tags/index.fr.html
@@ -20,7 +20,7 @@ f.status = 'published';  // should be one of draft, review, published, notreview
 f.path = '../../' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2023-08-29', time:'19:50'} // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2023-10-24', time:'16:20'} // date and time of latest edits to this document/translation
 f.contributors = ''; // people providing useful contributions or feedback during review or at other times
 // also make sure that the lang attribute on the html tag is correct!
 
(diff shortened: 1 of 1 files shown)


r12a:
hi @clavoline it took me a while to finish reviewing this because there are many differences in the way you arranged the text, but i think it looks ok, with one small exception (compared with the proposed new [english version](https://w3c.github.io/i18n-drafts/articles/language-tags/index.en.html)).

The note just below the top heading doesn't follow our current style.  You need to change the class from `note` to `info`.

Otherwise, i think we're good. Thanks.

clavoline:
@r12a Thank you. Sorry I missed that one!

r12a:
Ah, that didn't quite work.  I think we need to replace the `div` with a `p`.

r12a:
Looks good. Happy for me to publish and announce it ?

clavoline:
@r12a Of course, if you're happy with it!

r12a:
For the record, i had to fix a number of validation errors, and put section tags around all the new section headings (with ids).  The diff can be seen at https://github.com/w3c/i18n-drafts/commit/b026502df67a39600a6a0498adcbc2d4e6339243

clavoline:
@r12a My apologies - I forgot about the translation checklist and didn't realize sections were also required for H3s.

r12a:
np @clavoline   I just pointed to it so that you are aware for next time :-)

All announced now.
https://w3c.social/@webi18n/111341218907831498
https://twitter.com/webi18n/status/1720072602751029600

r12a:
@clavoline we have some feedback on mastodon, suggesting that étiquettes should be balises.  Do you think we should change it?  https://w3c.social/@webi18n/111341218907831498

r12a:
I think balise refers to markup tags, rather than language tags??

clavoline:
> I think balise refers to markup tags, rather than language tags??

Agreed. This is one of the first questions I asked myself!

These 3 sources use "étiquettes":
- http://hapax.qc.ca/CEISIC-Colloque2008-Andries-Indiquer-la-langue-version-2008.pdf
- http://abcdrfc.free.fr/rfc-vf/rfc4646.htm
- https://fr.wikipedia.org/wiki/%C3%89tiquette_d%27identification_de_langues_IETF

PR #561, opened 2023-11-13
Remove the mention of unicode-xml

xfq:
The document has been withdrawn.

[Preview](https://deploy-preview-561--i18n-drafts.netlify.app/questions/qa-chars-vs-markup.en.html)

### DIFF
--- a/questions/qa-chars-vs-markup.de.html
+++ b/questions/qa-chars-vs-markup.de.html
@@ -228,7 +228,7 @@ f.additionalLinks = ''
     <li>
       <p>Verwandte Links: <cite>XML erstellen</cite></p>
       <ul><li><a href="/International/techniques/authoring-xml#charset">Zeichen</a></li>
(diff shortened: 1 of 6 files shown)


r12a:
@xfq Are you planning to make changes to the final section about compatibility characters?  (I think you should.)

We can remove the sentence: "This is taken from [Unicode in XML & Other Markup Languages](http://www.w3.org/TR/unicode-xml/):".  But we should also carefully consider the (previously) quoted text, especially in light of the fact that we don't normally recommend converting these characters to their compatibility equivalents.

The table also needs to be thought about...  and probably significantly changed.

xfq:
> We can remove the sentence: "This is taken from [Unicode in XML & Other Markup Languages](http://www.w3.org/TR/unicode-xml/):".

Done. Thanks.

> The table also needs to be thought about... and probably significantly changed.

Do you have any specific suggestions?

Issue #564, opened 2023-12-04
[articles/idn-and-iri/index]  Information in the multilingual web addresses article needs to be updated


xfq:
[source] (https://www.w3.org/International/articles/idn-and-iri/) [en]

Some information in this article needs to be updated, like:

1. Only URI (RFC3986) and IRI (RFC3987) are mentioned in the article. We might want to add information about the WHATWG URL Standard.
2. We should update the HTML 4.0 example to "HTML".
3. We should update the links to the RFC specifications to point to https://www.rfc-editor.org/
4. "top level domains" should be "top-level domains"

>  In this case, if we were to use percent-escaping to transform the (same) characters in the address so that they to conform to the URI requirements, we would base the escapes on the bytes that represent 引き割り.html in Shift-JIS.

5. "they to conform to" above should be "they conform to".
6. `mod_fileiri` looks unmaintained, should we keep the reference to it?
7. The reference to Internet Explorer and Netscape should probably be removed.

> You can run a basic check to see whether IDNs work on your system using this simple test.

8. ^ There should be a more up-to-date test.

> Different browsers use different strategies to determine whether the URI should be shown in Unicode or punycode.

9. "URI" should be "IRI" instead above?
10. The handling of IDNs by different browsers is mentioned, but we should link to some more updated resources like https://chromium.googlesource.com/chromium/src/+/main/docs/idn.md and https://wiki.mozilla.org/IDN_Display_Algorithm#Algorithm

> There is a similar issue with the use of simplified vs. traditional characters in the Chinese Han script.

11. This isn't a huge problem, because if a character isn't unified, most people who know Simplified Chinese or Traditional Chinese can easily see the difference. The bigger problem are things like Kangxi radicals (such as U+2F04 乙 and U+4E59 乙) and duplicate encoded characters (such as 㘽 U+363D and 㦳 U+39B3), because the glyphs are often the same. Also, some registries solves this by making the simplified and traditional characters equivalent (see [1](https://www.cscdbs.com/blog/traditional-and-simplified-chinese-versions-of-the-internationalized-domain-name-idn-cctld-for-taiwan-now-part-of-tw-bundle/) and [2](https://www.cnnic.net.cn/n4/2022/0919/c90-10605.html)).

> There are some improvements needed to the specifications for IDN and IRIs, and these are currently being discussed. For example, there is a need to extend the range of Unicode characters that can be used in domain names to cover later versions of Unicode, and to allow combining characters at the end of labels in right to left scripts.

[...]

PR #565, opened 2023-12-08
Update the multilingual web addresses article

xfq:
Partial fix for #564.

This PR fixes a few simpler issues mentioned in #564, including:

- updating RFC URLs
- updating the links to the RFC specifications to point to https://www.rfc-editor.org/
- "top level domains" -> "top-level domains"
- removing the mention of the unmaintained `mod_fileiri`
- removing the reference to Internet Explorer and Netscape
- updating some links

After merging this, I'll work on adding things like WHATWG URL, IDN in browsers,  IDNA2008/UTS46, and maybe write some new tests.


r12a:
Just a note on the use of the `modifiers` variable: this shouldn't be used for just editorial changes, such as fixing URLs or typos. Its intended use is for situations where major substantive changes have been applied to the article — which is why it is generally not commonly used.  

Fwiw, i also tend not to credit contributors who point out editorial issues; only those who make substantive contributions are listed.

xfq:
OK, I removed myself from the `modifiers` variable. Thanks.

PR #570, opened 2023-12-15
Translate qa-bidi-controls into zh-hans

### LINKED ISSUE #568: [questions/qa-bidi-controls]  i18n-html-tech-bidi is outdated
[source] (https://www.w3.org/International/questions/qa-bidi-controls) [en]

> Using escapes is also problematic, since the characters that make up an escape can be mangled by editors in bidirectional text. See [Authoring HTML: Handling Right-to-left Scripts](https://www.w3.org/TR/2014/NOTE-i18n-html-tech-bidi-20140603/#bidisource) for more information.

Perhaps we should link to https://www.w3.org/International/questions/qa-bidi-source.en.html instead.

xfq:
Translate *Unicode controls vs. markup for bidi support* into zh-hans and fix #568.

[Preview](https://deploy-preview-570--i18n-drafts.netlify.app/questions/qa-bidi-controls.zh-hans.html)


xfq:

PR #581, opened 2024-03-30
Initial work on language-negotiation materials

aphillips:
I started to convert some non-prose (powerpoint) presentation materials to prose describing language negotiation. This PR is not ready to merge, as more material needs to be added and the text given a thorough edit. However, it is provided for commentary.


annevk:
I think this needs to include analysis about the potential privacy implications. In particular for minority end users.

In some sense, complete customization based on the end user's preferences is rightly an ideal, but that needs to be weighed against websites being largely unvetted applications that might not have the end user's best interests in mind.

aphillips:
> I think this needs to include analysis about the potential privacy implications. In particular for minority end users.

@annevk Thanks for the comment.

This material was written from the point of view of a website owner/developer, so, of course, doesn't currently include such considerations. That said, it probably should have already included some admonition about not exposing user preferences to third parties.

I agree that privacy implications need to be addressed and will add some text about that shortly. Comments/suggestions welcome.

xfq:
The scope of this article seems to have some overlap with our [existing article](https://www.w3.org/International/questions/qa-when-lang-neg.en.html), both covering when language negotiation should be used. The existing article contains more examples.

Should we probably introduce the concept of language negotiation in this article, and introduce when we should use language negotiation in our existing article?

xfq:
This document seems to discuss both "language negotiation" and "locale negotiation". If these two terms have different meanings, it might be useful to point them out clearly so that readers will feel clearer. If not, it would be best if we could use consistent wording.

aphillips:
> This document seems to discuss both "language negotiation" and "locale negotiation". If these two terms have different meanings, it might be useful to point them out clearly so that readers will feel clearer. If not, it would be best if we could use consistent wording.

These terms _don't_ have different meanings. The problem is that "language negotiation" is the common term, while "locale negotiation" is more accurate.

It's a good callout that I don't define the terms and pull them together early in the document. See 5950acd for fixes.

articles/language-negotiation/language-negotiation.md
```diff
@@ -203,26 +218,29 @@ outside that region.
...
-> 2. If the user's geographic region has a default let return value be that language.
+> 2. If the user's geographic region has a default let return value be that locale.
 > 3. If the user has an Accept-Language header
->    i. For each language in the A-L header
->       a. if the language matches an available language, let return value be that language
-> 4. If the user has a cookie with a language preference, let return value be that language
-> 5. If the URL contains a language, let return value be that language
-> 6. If the user is recognized, let return value be the language in the user's profile
+>    i. For each language range in the A-L header
+>       a. if the language range matches an available locale's language tag, let return value be that locale
```
xfq: nit: Markdown doesn't seem to support list syntax like `i.` or `a.`.

aphillips: Yes: the article will be converted to HTML before final publication.

klensin:
Sorry... long overdue and most of this is nitpicking.  I'm trying to transcribe notes from about 10 days ago rather than letting it go longer; hope the result is sufficiently clear.

(i) Probably the first use of "language tag" should be a link to BCP 47.  Please do not link just to the registry
(ii) 2nd paragraph, first example or earlier: mention impact on trying to compare one instance of a string with might be another instance.
(iii) Bullet list "Determining...".  IP addresses are increasingly unreliable for determining locations or much of anything else other than the address itself  The problems arise from efforts to deal with the shortage by NAT mapping before the destination LAN (e.g., with CGNAT), assorted 4to6 and 6to4 mappings), and assorted privacy-driven efforts.  No need to go into that in the document, just be careful about mention of determining local from addresses.
(iv) Text about the simplest case, "persist" should probably be "preserve".
(v) The last section is really not clear.  In particular, the user should probably be given an explicit choice.

articles/language-negotiation/language-negotiation.md
```diff
@@ -0,0 +1,340 @@
...
+
+A [locale](https://www.w3.org/TR/i18n-glossary/#dfn-locale) is:
+> An identifier (such as a language tag) for a set of international preferences. 
+> Usually this identifier indicates the preferred language of the user and 
+> possibly includes other information, such as a geographic region (such as a country). 
+> A locale is passed in APIs or set in the operating environment to obtain culturally-affected behavior within a system or process.
+
+Developers pass the locale (or set it in the operating environment) when calling 
+internationalized APIs used for operations such as loading the localized resources mentioned above.
+But they are also use APIs that make the content or data in a system appear or
```
xfq: 'they are also use' -> 'they also use'?

PR #588, opened 2024-06-13
Update pointer to review radar

aphillips:


aphillips:
Need to update this page also, but can't find the source??

https://www.w3.org/International/review-request

New target: https://github.com/orgs/w3c/projects/91

xfq:
> Need to update this page also, but can't find the source??
> 
> https://www.w3.org/International/review-request

It's on CVS. I'll update this page.

xfq (approved):

PR #607, opened 2024-11-06
Change the title of alreq

xfq:
From Arabic & Persian to Arabic Script languages.


r12a:
Ah, we need to be careful here.  The name of the lreq file is indeed Arabic Script Resources, but the layout requirements document we have is solely focused on Arabic & Persian, so it has retained its title as such.  In the future we may have an Urdu & Kashmiri Layout Requirements doc, or a Uighur one, etc. since these are quite different orthographies.  They'll still fall under the same lreq doc (since that is related to the script).  I don't have time to check the context of the various documents you propose changes for – could you do that?

xfq:
All the places I modified refer to the group (task force), not the name of any specific document.

r12a (approved):

PR #616, opened 2024-11-25
Add lang attribute for Thai text

xfq:


xfq:

PR #622, opened 2024-11-28
Address formats around the world

xfq:
Started to write some stuff about address formats. Comments are welcome.

-----

If this is published, we might want to consider updating [specdev](https://w3c.github.io/bp-i18n-specdev/#locale) and the [techniques](https://www.w3.org/International/techniques/authoring-html) page.

[Preview](https://deploy-preview-622--i18n-drafts.netlify.app/questions/qa-address-formats.en.html)


questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,107 @@
+# Address formats around the world
+
+## Question
+
+### How do address formats differ around the world, and what are the implications of those differences on the design of forms, databases, ontologies, etc. for the Web?
+
+Address formats vary widely across the globe, with differences in structure, content, and the level of granularity. For authors and developers designing forms, databases, or systems that handle addresses, understanding these variations is crucial to avoid frustrating users from other countries. This article will introduce some of the key differences in address formats around the world and provide guidance on how to design systems that can handle them effectively.
+
+This is not an exhaustive guide but aims to sensitize you to the complexities of international address formats and the challenges they pose for web design. As with [personal names](https://www.w3.org/International/questions/qa-personal-names.en.html), there is rarely a "perfect" solution, but awareness of these differences is the first step toward building more inclusive systems.
```
r12a: Don't use the .en.html in the link, since we have translations of that article.

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,107 @@
...
+
+Elsewhere in Japan, street names are often not used at all. Instead, buildings are referenced by block numbers, and their numbering may not follow geographic order (e.g., they can be the order in which buildings were constructed).
+
+#### **Order of address components**
+
+While most countries recommend that users format addresses in order from the most detailed part (such as street and house number) to the most general part (like city and province), there are some cases that slightly deviate from the general rule.
+
+In Germany, if there are suburbs, their names should be placed above the street name and house number.
+
+In China, addresses are written starting with the postal code, followed by the largest administrative area (e.g., province), and down to the smallest unit (e.g., room number). For example:
```
r12a: and Japan

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,107 @@
...
+
+```
+Russia, Moscow, Tverskaya Street, 16
+```
+
+However, Russia also uses the "smallest-to-largest" order commonly seen internationally, so developers should account for both formats.
+
+#### **Locality-specific elements**
+
+For example, the UK includes elements like "dependent locality" and "double dependent locality" for more granular location information. Royal Mail requires the "post town" to be included whenever possible.
```
r12a: Even though i live in the UK i don't understand what this dependency stuff means ;-). Perhaps reword, explain, or give an example of what this is about.

questions/qa-address-formats.en.md
r12a: A difference that i think is worth mentioning is the order of street name and house number.  In Germany & Switzerland the house number follows the street name, which has implications for form design.

Also, some people may not have house numbers, or may need to add a house name, or a flat number.  This leads to the more general issue, too, that the number of lines needed for an address can vary.

There are also things like where the address should be printed on an envelope - in Switzerland i used to get in trouble for putting the 'to' address in the middle.

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,107 @@
...
+If your website serves users from multiple countries, consider localizing the address form for each locale, such as displaying field labels in the user's language, adjusting the order of fields to match the address format of the user's country, and using address auto-completion tools that are aware of regional differences.
+
+## Further reading
+
+For further reading on related topics, see:
+
+- [International Address Format: Structure and Normalization](https://www.geopostcodes.com/blog/international-address-data/)
+- [Address](https://en.wikipedia.org/wiki/Address) on Wikipedia
+- [Proposal for extending the autocomplete attribute](https://github.com/battre/autocomplete-attribute-explainer/)
+- [Personal names around the world](https://www.w3.org/International/questions/qa-personal-names.en.html)
```
r12a: Again, remove .en.html

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,107 @@
+# Address formats around the world
+
+## Question
+
+### How do address formats differ around the world, and what are the implications of those differences on the design of forms, databases, ontologies, etc. for the Web?
+
+Address formats vary widely across the globe, with differences in structure, content, and the level of granularity. For authors and developers designing forms, databases, or systems that handle addresses, understanding these variations is crucial to avoid frustrating users from other countries. This article will introduce some of the key differences in address formats around the world and provide guidance on how to design systems that can handle them effectively.
```
r12a: I think it would be good to include lots of visual example, pretty much for each point you make, rather than just the few you have.

r12a (approved):

questions/qa-address-formats.en.md
xfq: Thanks, I added more examples.

questions/qa-address-formats.en.md
xfq: Added.

questions/qa-address-formats.en.md
xfq: I decided to remove this paragraph, because ordinary people don't need to know this, only address experts need it :)

questions/qa-address-formats.en.md
xfq: Updated.

questions/qa-address-formats.en.md
xfq: Updated.

questions/qa-address-formats.en.md
xfq: Do you have any examples for the points you mentioned? I can put them in the article when I write.

r12a: German example (using [this generator](https://capitalizemytitle.com/name-generator/german-street/)):

> Sternengasse 3
> 35108 Allendorf
> Deutschland

Example of complicated street address:

> Flat 4, Danson House
> 63 Somerton Road
> Market Drayton
> UK

xfq: Thanks. Added.

r12a: i think the UK example is worth mentioning by explaining that the street name and house number are not necessarily on the first line in an address. It may be necessary to indicate the appartment number and building/house name first.  In fact, that may be preceded by ℅ (care of) and a person's or organisation's name too. The user must have the flexibility to add these other lines, as needed.  This is then picked up again later when you offer advice to use field names like 'Line 1' etc, rather than describing what the field contains.

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,131 @@
...
+```
+100083 北京市 海淀区 学院路 37号
+```
+
+(Postal code → City → District → Street → House number).
+
+Russia often follow a "largest-to-smallest" order, starting with the country. For example:
+
+```
+Russia, Moscow, Tverskaya Street, 16
```
r12a: Especially because you used native Chinese & Japanese text examples just above, i think this needs to be in cyrillic script.  The transcription with arrows below the Chinese works well, and could be added after the Russian, too.

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+However, Russia also uses the "smallest-to-largest" order commonly seen internationally, so developers should account for both formats.
+
+#### **Countries without postal codes**
```
r12a: I'd move this section below Postal Code Systems Aroudn the World

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,131 @@
...
+
+Not all countries use postal codes. For example, Angola and some other nations do not use postal codes, which can make validation more challenging. Always design systems with flexibility to accommodate such exceptions.
+
+## Postal code systems around the world
+
+Postal codes also vary widely in format and granularity:
```
r12a:
```suggestion
Postal codes vary widely in format and granularity:
```

questions/qa-address-formats.en.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Length and format
```
r12a: while reading this i'm learning about lengths, but i'm asking myself which patterns include multiple groupings, like the uk - eg. are Argentina 8 digit codes a single string, or do they have groups that are separated by spaces or hyphens?  Note btw, that the separator is important — UK post codes are never written with dashes between the parts

r12a (changes requested):

questions/qa-address-formats.en.md
xfq: Added the UK example now.

questions/qa-address-formats.en.md
xfq: Fixed.

questions/qa-address-formats.en.md
xfq: Done.

questions/qa-address-formats.en.md
xfq: Fixed.

questions/qa-address-formats.en.md
xfq: Fixed by adding more information about postal codes.

r12a:
All change LGTM. This is looking quite good now.

questions/qa-address-formats.en.md
```diff
@@ -143,7 +143,7 @@ Not all countries use postal codes. For example, Angola and some other nations d
 
 **Provide additional delivery instructions.** Include dedicated fields for delivery instructions or notes that can accommodate informal addressing information. This might include fields like "Delivery Notes".
 
-**Allow non-ASCII characters.** Many addresses include characters from non-Latin scripts (e.g., Chinese, Japanese, Cyrillic). Use a Unicode encoding (such as UTF-8) in your pages, backend databases, and all software code in between.
+**Allow non-ASCII characters.** Many addresses include characters from non-Latin scripts. Use ASCII addresses for international shipping labels and customs forms, but include native script addresses as supplementary information when possible. Provide native script addresses to local delivery services for the final delivery stage.
```
r12a: I'm not sure it needs to be ASCII.  I think Latin text is ok.  For example, you don't normally need to write Spanish addresses without accents.

xfq: fixed

questions/qa-address-formats.en.html
```diff
@@ -0,0 +1,183 @@
...
+
+  <h1>Address formats around the world</h1>
+</header>
+
+
+  <div id="audience">
+    <div id="updateInfo"></div><script>document.getElementById('updateInfo').innerHTML = g.updated</script>
+  </div>
+
+<section id="question">
```
r12a, line 57: The structure of the sections appears to be badly formed.  See the TOC.

questions/qa-address-formats-data/local.css
```diff
@@ -0,0 +1,22 @@
+.highlight {
```
r12a: I'm thinking that at least some of these styles should perhaps be defined in article-2022.css.

r12a (changes requested):

questions/qa-address-formats.en.html
xfq: Thanks. Should be fixed now.

questions/qa-address-formats-data/local.css
xfq: Indeed. I just did that.

PR #624, opened 2024-11-29
Add Bluesky

xfq:


r12a (approved):
I can't see the result in preview, but assuming that you tested it LGTM.  We'll need to change it on the home page in CVS, too.

aphillips (approved):
LGTM

PR #634, opened 2025-02-14
Update qa-indic-graphemes.en.html

r12a:
Takes into account the new rules for grapheme clusters when conjuncts are involved.

[Preview](https://deploy-preview-634--i18n-drafts.netlify.app/questions/qa-indic-graphemes.en.html)


r12a:
Reviewed during i18n telecon, with decision to go ahead and publish changes.

PR #647, opened 2025-05-04
feat(i18n): Translate qa-lang-why into Persian

Goudarz:
[Preview](https://deploy-preview-647--i18n-drafts.netlify.app/questions/qa-lang-why.fa.html)

### DIFF
--- a/questions/qa-lang-why.fa.html
+++ b/questions/qa-lang-why.fa.html
@@ -1,0 +1,169 @@
+<!DOCTYPE html>
+<html lang="fa-IR" dir="rtl">
+<head>
+<meta charset="utf-8">
+<title>چرا از ویژگی زبان استفاده کنیم؟</title>
+<meta name="description" content="چرا باید از ویژگی زبان در برگه‌های وب استفاده کنم؟">
+<script>
+var f = { }
+
+// AUTHORS should fill in these assignments:
+f.directory = 'questions'+'/' // the path to this file, not including /International or the file name
+f.filename = 'qa-lang-why' // the file name WITHOUT extensions
+f.authors = 'Richard Ishida, W3C' // author(s) and affiliations
+f.previousauthors = 'Deborah Cawkwell, BBC World Service' // as above
+f.modifiers = '' // people making substantive changes, and their affiliation
+f.searchString = 'qa-lang-why' // blog search string - usually the filename without extensions
+f.firstPubDate = '2004-06-21' // date of the first publication of the document (after review)
+f.lastSubstUpdate = { date:'2014-11-18', time:'11:00'}  // date and time of latest substantive changes to this document
+f.status = 'published'  // should be one of draft, review, published, notreviewed or obsolete
+f.path = '../' // what you need to prepend to a URL to get to the /International directory 
+
+// AUTHORS AND TRANSLATORS should fill in these assignments:
+f.thisVersion = { date:'2025-05-04', time:'20:59'} // date and time of latest edits to this document/translation
+f.contributors = '' // people providing useful contributions or feedback during review or at other times
+// also make sure that the lang attribute on the html tag is correct!
+f.sources = '' // describes sources of information
+
+// TRANSLATORS should fill in these assignments:
+f.translators = '<a href="https://goudarzjafari.com" title="Goudarz Jafari">Goudarz Jafari</a>' // translator(s) and their affiliation - a elements allowed, but use double quotes for attributes
+
+f.breadcrumb = 'language'
+
+f.additionalLinks = ''
+</script>
+<script src="qa-lang-why-data/translations.js"> </script>
+<script src="../javascript/doc-structure/article-dt.js"> </script>
+<script src="../javascript/boilerplate-text/boilerplate-fa.js"> </script>
+<!--TRANSLATORS must change -en in the line just above to the subtag for their language! -->
+<script src="../javascript/doc-structure/article-2022.js"> </script>
+<script src="../javascript/articletoc-2022.js"></script>
+<link rel="stylesheet" href="../style/article-2022.css">
+<link rel="copyright" href="#copyright">
+<script src="../javascript/prism.js"></script>
+<link rel="stylesheet" href="../style/prism.css">
+</head>
+
+<body>
+<header>
+  <nav id="mainNavigation"></nav><script>document.getElementById('mainNavigation').innerHTML = mainNavigation</script>
+  
+  <h1>چرا از ویژگی زبان استفاده کنیم؟</h1>
+</header>
+
+
+  <div id="audience">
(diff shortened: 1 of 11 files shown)


xfq:
Not sure if we should use `fa` or `pes` as the language tag here. Any guidance @aphillips @r12a?

aphillips:
Probably fa. Cldr uses that subtag and it's the most common in the wild I believe.

xfq:
@r12a's Persian page seems to use `pes`, though: https://r12a.github.io/scripts/arab/pes.html

r12a:
Yes, i decided to use `pes` for my orthography notes, after a little deliberation.

`fa` is certainly more widespread afaik, but it is a macrolanguage (see https://r12a.github.io/app-subtags/?find=persian), and i thought that i might, some time in the future, want to produce an orthography notes page for Dari (as opposed to Iranian Persian), therefore i chose the more specific label.

That said, a lot of people assume that `fa` is the language tag for Persian.  If they are likely to search for content using that label, it might be worth using it, but i'm not sure whether that would really apply here.  We mostly find content in a particular language by following links.

Note, also, that we do generally recommend using the more specific labels at https://www.w3.org/International/questions/qa-choosing-language-tags.en.html#langsubtag

And the labels in the LE framework also use `pes`.

So, personally, i'd go with `pes`.

r12a:
I also suspect that using `pes` will simplify our file structure and translation database, because you won't need to distinguish both `fa` and `fa-IR` in a number of the pull request submissions.

xfq:
I also prefer `pes` for the same reasons as @r12a, and I don't see any disadvantages to using `fa` so far. @Goudarz Can you update the language tag to `pes`?

questions/qa-lang-why.fa.html
```diff
@@ -0,0 +1,169 @@
...
+<h2>جزئیات</h2>
+<p>ما در اینجا چند روش که در حال حاضر برای اطلاعات زبانی مفید هستند را فهرست کردیم، اما با پیشرفت استانداردها و مرورگرها در آینده، کاربردهای بیشتری برای اطلاعات زبانی ممکن خواهد شد.</p>
+  
+  <section id="rendering">
+<h3>ظاهر برگه‌ها</h3>
+    <p>ویژگی‌های زبان به شما این امکان را می‌دهند که ظاهر محتوای خود را بر اساس زبان تغییر دهید. برای اطلاعات بیشتر درباره‌ی نحوه انجام این کار، به <a class="print" href="/International/questions/qa-css-lang"><cite>ظاهر دادن با استفاده از ویژگی lang</cite></a> مراجعه کنید.</p>
+    <p>به‌عنوان مثال، قلم‌ها یا فاصله‌ی خطوط ممکن است برای هماهنگی با الفباهای مختلف تغییر کنند، علامت‌های نقل‌قول تولید‌شده بر اساس ظاهر ممکن است نیاز به تغییر بسته به زبان داشته باشند، و ممکن است به روش‌های وابسته به زبان بیان شود و غیره.</p>
+    <p>مثال زیر نشان می‌دهد که چگونه می‌توان یک قلم خاص را برای متن عربی جاسازی‌شده در یک برگه تنظیم کرد.</p>
+    <figure class="example" translate="no">
+      <pre><code class="language-css">body {
```
xfq, line 84: We need to add `dir=ltr` here to make the code render correctly.

r12a:
@Goudarz thank you for the translations!  @xfq as part of the work i've been doing this week, i plan to close this PR.  I have taken the translations that are appropriate and made the changes needed, and will commit them shortly.  The changes include:
1. validation corrections
2. fa -> pes (and remove fa-IR)
3. apply the new template for the boilerplate-pes file
4. ignore changes to legacy files that will never need a Persian translation
5. fix the code block example

@Goudarz it's great that you figured out all the connections, and translated all the bits we normally have to change for a new translation language.  If you do more translations you'll only need to translate the text in the main page, now.  However, bear in mind that it's always best to ping us before starting the translation – we will usually apply new template changes and fixes to the source before you start working on it, so that we don't need to fix them twice. Cheers.

r12a:
@Goudarz @xfq please check out https://w3c.github.io/i18n-drafts/questions/qa-lang-why.pes.html

If everything looks ok, we can copy to the CVS server and announce.

r12a:
@Goudarz would you like the slanted text next to the W3C icon at the top to slant towards the left?
eg.

Screenshot 2025-05-10 at 14 11 18

(We could, in fact, make that happen for `em` text in the body of the content, too, though that might be a step too far.)

Goudarz:
> @Goudarz @xfq please check out https://w3c.github.io/i18n-drafts/questions/qa-lang-why.pes.html
> 
> If everything looks ok, we can copy to the CVS server and announce.

It looks good and ready to announce. 

> @Goudarz would you like the slanted text next to the W3C icon at the top to slant towards the left? eg.
> Screenshot 2025-05-10 at 14 11 18
> 
> (We could, in fact, make that happen for `em` text in the body of the content, too, though that might be a step too far.)

I believe left slant for `em` text is correct. It might seem like an extra step but I think it would be necessary.

xfq:
Looks good to me. @r12a  are you going to publish the translation, or should I do it?

r12a:
Thanks again @Goudarz !  

Files now uploaded to the W3C server at https://www.w3.org/International/questions/qa-lang-why.pes.html

Announced at BlueSky and Mastodon: https://bsky.app/profile/webi18n.bsky.social/post/3lp7tvk75q225

r12a:
Btw, note that the tagline at the top of the page and the `cite` styling are all styled to lean to the left (the direction of reading).  Unfortunately, that is only supported by the Gecko engine at the moment (we have raised it as a gap).

Screenshot 2025-05-15 at 16 56 35

 This was done by using `font-style: oblique -14deg;`  It's not as good as using a proper left-leaning italic font, but the results are reasonably good.

aphillips:
Following up on the `fa`/`pes` discussion [above](https://github.com/w3c/i18n-drafts/pull/647#issuecomment-2862317403).

`fa` is a macrolanguage, much like `zh` is a macrolanguage. The practice in CLDR is to associate the "dominant" variety with the macrolanguage (when a dominant variety exists). That's why we usually see tags like `zh-Hans` or `zh-CN` rather than `cmn-Hans` or `cmn-CN` for Chinese: this does not mean that the `cmn` tags are "wrong". But they are generally frowned upon by comparison.

And it's the reason that the locales are `fa-*` rather than being `pes-*` (with Dari being consistently tagged `prs`). Preferring `fa` for locales does not make the `pes` tags wrong. However, people might find it confusing that Unicode and W3C do things differently ("where's `fa`?" might be a question, given that there are locales for it in e.g. `Intl.Locale`)

PR #651, opened 2025-05-11
Update boilerplate-hu.js

iherman:
Few comments on the changes:

- I used the English boilerplate in case of doubt. I hope that is all right.
- I was not 100% what goes where on the final text. The problem that may arise is that Hungarian doesn't use the English capitalization rules in titles, headings, etc. Ie, only the first character is capitalized and not all the words. I kept the capitalization here and there, but that may have to be changed depending on its final use.
- I would strongly advise getting this reviewed by someone who is not only native Hungarian like me, but also _lives_ in Hungary. I left Hungary before the appearance of the Web and, though I regularly read Web pages in Hungarian, so I may have got some expressions (Web site, RSS feed...) wrong.
- I also tried to avoid having to make a choice between the formal style ("vouvoiement" in French) and the informal style but I could not do that everywhere. The problem is that my instincts in this respect might be strongly outdated in terms of Web use, I believe that has changed a lot in Hungary. Again, getting someone who lives in Hungary (and who is younger than me) may be a good idea.

With all these caveats, I hope this could help...

### DIFF
--- a/javascript/boilerplate-text/boilerplate-hu.js
+++ b/javascript/boilerplate-text/boilerplate-hu.js
@@ -28,7 +28,7 @@ s.currLang = {
   'en':'Angol',
   'es':'Spanyol',
   'fr':'Francia',
-  'gl':'Gallego',
+  'gl':'Galiciai',
   'he':'Héber',
   'hi':'Hindi',
   'hu':'Magyar',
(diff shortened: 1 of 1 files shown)


r12a:
Wow. Lot's of changes.  Many thanks, Ivan !

r12a (approved):

PR #652, opened 2025-05-12
Update boilerplate-nl.js

bert-github:
Added some missing translations and changed a few old ones for consitency: use infinitive for menu commands, use ‘homepage’ instead of ‘thuispagina’, etc.


r12a:
Thanks very much, Bert !  I'll merge and process it...

PR #654, opened 2025-05-14
feat(i18n): Translate uba-basics into Persian

Goudarz:
[Preview](https://deploy-preview-654--i18n-drafts.netlify.app/articles/inline-bidi-markup/uba-basics.pes.html)

### DIFF
--- a/style/article-2022.css
+++ b/style/article-2022.css
@@ -232,6 +232,8 @@ figcaption::before { content: "Fig. " counter(figure) ". "; }
 
 figcaption:lang(zh-hans)::before { content: "图" counter(figure) ":\00A0 "; }
 
+figcaption:lang(pes)::before { content: "تصویر " counter(figure, persian) ". "; }
+
 /* ************** set margins **************** */
 
 p, h2, h3, h4, dl, figure { margin-inline-start: 7.5%; margin-inline-end: 32%; }
@@ -268,6 +270,7 @@ figure p, figure pre { margin-inline-start: 0; margin-inline-end: 0; }
 .inlinedemolink { font-size: 80%; margin-inline-start: 1em; margin-start: 1em;  }
 figcaption { font-style: italic; }
 figcaption:lang(zh) { font-style: normal; }
+figcaption:lang(pes) { font-style: normal; display: inline-block; transform: skew(7deg); padding: 0 2px; -webkit-transform: skew(7deg); -moz-transform: skew(7deg); }
 
 dt, dd { margin-block-end: 0; }
 dd .example { margin-inline-start: 0; margin-inline-end: 0; }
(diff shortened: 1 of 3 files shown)


r12a:
@Goudarz you had asked about changing the images to suit Persian. Do you still plan to do that? You're welcome to replace the images with text translated into Persian. (Btw, if you do, it may help to follow the "See live demo." links and grab the images after modifying the HTML  — that will help maintain the look & feel across the pages.)

style/article-2022.css
```diff
@@ -268,6 +270,7 @@ figure p, figure pre { margin-inline-start: 0; margin-inline-end: 0; }
 .inlinedemolink { font-size: 80%; margin-inline-start: 1em; margin-start: 1em;  }
 figcaption { font-style: italic; }
 figcaption:lang(zh) { font-style: normal; }
+figcaption:lang(pes) { font-style: normal; display: inline-block; transform: skew(7deg); padding: 0 2px; -webkit-transform: skew(7deg); -moz-transform: skew(7deg); }
```
r12a: I see you're using the `transform` property, rather than `font-style oblique`.  That does work with more browsers, so perhaps we should use both until the browsers all support font-style properly? 

I also notice that you're using a 7deg skew here, but 10deg for the cite.  Was that intentional?  If so, i can replicate it in the font-style declarations.

Fwiw, i had used a 14deg slant just because that corresponds to the Latin script default in the Fonts spec.

style/article-2022.css
```diff
@@ -1859,5 +1876,12 @@ h2:hover > a.selflink, h3 > a.selflink, h4 > a.selflink, h5 > a.selflink, h6 > a
 	}
 /* END IN-DOC TESTS */
 
-
+em:dir(rtl) {
```
r12a: This sets the left lean for emphasis in all languages that are written RTL, which is not what we want (Adlam preference is right-leaning, and majority Hebrew is also).  Could you tailor this for pes only.

style/article-2022.css
```diff
@@ -232,6 +232,8 @@ figcaption::before { content: "Fig. " counter(figure) ". "; }
 
 figcaption:lang(zh-hans)::before { content: "图" counter(figure) ":\00A0 "; }
 
+figcaption:lang(pes)::before { content: "تصویر " counter(figure, persian) ". "; }
```
r12a, line 235: Nice!

r12a (changes requested):

r12a:
Btw, i'm aware that the examples in the images are written with the frame of reference of a LTR script person.  It would be possible to adapt the examples so that they look at the world from a RTL perspective.  However, that would be a fair bit more involved that just translating the text, so i don't see it as essential.

style/article-2022.css
r12a: Btw, i haven't uploaded my local version of the .css file (luckily) to GH. If i did, we'd have a conflict, so i'll hang back on that until we decide what to do here.

Currently my local file says things like:

````
cite:lang(pes) {
    font-style: oblique -14deg;
    }
````

Goudarz:
@r12a  Yes, I plan to translate the images and live demo into Persian to make them more accessible to the Persian-speaking community.

PR #655, opened 2025-05-19
Add more information about when to use (and not use) the BOM

xfq:
And mention UTF-32 in the document.

[Preview](https://deploy-preview-655--i18n-drafts.netlify.app/questions/qa-byte-order-mark.en.html)

### DIFF
--- a/questions/qa-byte-order-mark.en.html
+++ b/questions/qa-byte-order-mark.en.html
@@ -20,8 +20,8 @@ f.status = 'published'  // should be one of draft, review, published, notreviewe
 f.path = '../' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2016-04-20', time:'11:10'} // date and time of latest edits to this document/translation
-f.contributors = 'Albert Lunde, Asmus Freytag, Björn Höhrmann, Henri Sivonen, John Cowan, Leif Halvard Silli, Norbert Lindenberg, Gwendoline Clavé' // people providing useful contributions or feedback during review or at other times
+f.thisVersion = { date:'2025-05-19', time:'11:10'} // date and time of latest edits to this document/translation
+f.contributors = 'Albert Lunde, Asmus Freytag, Björn Höhrmann, Fuqiao Xue, Henri Sivonen, John Cowan, Leif Halvard Silli, Norbert Lindenberg, Gwendoline Clavé' // people providing useful contributions or feedback during review or at other times
 // also make sure that the lang attribute on the html tag is correct!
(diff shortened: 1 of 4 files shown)


questions/qa-byte-order-mark.en.html
```diff
@@ -104,7 +104,8 @@ <h2>Answer</h2>
 <h3> What is a byte-order mark?</h3>
 
 <div class="sidenoteGroup">
-<p>At the beginning of a page that uses a <a class="termref" href="/International/articles/definitions-characters/#unicode">Unicode</a> <a class="termref" href="/International/articles/definitions-characters/#charsets">character encoding</a> you may find some bytes that represent the Unicode code point U+FEFF BYTE ORDER MARK (abbreviated as <dfn>BOM</dfn>).</p>
+<p>A <dfn>Byte Order Mark (BOM)</dfn> is a special <a class="termref" href="/International/articles/definitions-characters/#unicode">Unicode</a> character (U+FEFF) that can appear at the very beginning of a text file. Its primary original purpose was to indicate the byte order (or <a href="https://en.wikipedia.org/wiki/Endianness">endianness</a>) for 16-bit and 32-bit Unicode <a class="termref" href="/International/articles/definitions-characters/#charsets">encodings</a>. While often invisible and intended to aid in correctly interpreting text, the BOM can sometimes cause unexpected display issues or problems with software if not handled correctly, particularly with UTF-8 encoded files where its use is not for endianness.</p>
```
aphillips: A lot of things seem crammed into one paragraph. Plus there might be some missing info? Perhaps:

```suggestion
<p>A <dfn>Byte Order Mark</dfn>, sometimes abbreviated "BOM", is a special Unicode character intended to appear at the very beginning of a text file. Its original purpose was to indicate the <q><a href="https://en.wikipedia.org/wiki/Endianness">endianness</a></q> of text that used the UTF-16 or UTF-32 character encodings of Unicode. The Byte Order Mark is <kbd>U+FEFF ZERO WIDTH NON-BREAKING SPACE</kbd>: the character name refers to a separate, deprecated, use of the character.</p>
<p>Some systems use the BOM code point at the start of a file to indicate that text files are using the UTF-8 character encoding, even though UTF-8 does not need a marker to indicate endianness.</p>
<p>While often invisible and intended to aid in correctly interpreting text, the presence of the BOM can sometimes cause unexpected display issues or problems with software if not handled correctly.</p>
```

aphillips (changes requested):

questions/qa-byte-order-mark.en.html
xfq: Thanks. I adjusted the opening text.

aphillips (approved):
Thanks!!

PR #660, opened 2025-05-20
Tweak the About Page

xfq:
If it's OK to go, I'll update the translation as well.

### DIFF
--- a/nav/about.html
+++ b/nav/about.html
@@ -2,8 +2,8 @@
 <html lang="en">
 <head>
 <meta charset="utf-8">
-<title>About W3C i18n</title>
-<meta name="description" content="Description of the W3C Internationalization Activity.">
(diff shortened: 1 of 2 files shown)


nav/about.html
```diff
@@ -87,7 +87,7 @@ <h2>What is Internationalization?</h2>
 
 
 <section id="codebase">
-<h3>A universal code base</h3>
+<h3>A universal code base for multilingual applications</h3>
```
r12a: I would strongly advise against this, since it makes it appear that the following text can be skipped if you're not developing a multilingual application — which many developers _will_ think. But in fact, it's important to have a universal code base for all applications, whether or not you plan to make your application multilingual.  This is also relevant for web page design, not just applications.

xfq: Indeed, I reverted this line. Thanks.

PR #664, opened 2025-05-23
Translate qa-apache-lang-neg into zh-hans

xfq:
[Preview](https://deploy-preview-664--i18n-drafts.netlify.app/questions/qa-apache-lang-neg.zh-hans.html)


r12a:
I was thinking we should add text-autospace to the style sheet for Chinese.  At least it may work for macOS and iOS users.

xfq:
I thought about it, but the default value of this property is to apply the extra spacing. It's just [WebKit has not yet implemented it](https://github.com/WebKit/WebKit/pull/40309), so I think it may be a bit redundant...

xfq:
I'll add it. We can remove it after major browsers implement it.

xfq:

PR #665, opened 2025-05-23
Try redesigning the style of authoring-html

xfq:
Try redesigning the style of Internationalization techniques: Authoring web pages. Let me know what you think!

Current page: https://www.w3.org/International/techniques/authoring-html

[Preview](https://deploy-preview-665--i18n-drafts.netlify.app/techniques/authoring-html.en.html)

### DIFF
--- a/techniques/authoring-html.en.html
+++ b/techniques/authoring-html.en.html
@@ -61,6 +61,24 @@ div.content {
     float: none;
 }
 
+/* Responsive fix for narrow screens */
+@media (max-width: 768px) {
+    div.content {
+        margin: 1em;
+        padding: 1.5em;
+        max-width: none;
+        box-sizing: border-box;
+    }
+}
+
+@media (max-width: 480px) {
+    div.content {
+        margin: 0.5em;
+        padding: 1em;
+        border-radius: 4px;
+    }
+}
+
 /* Headings */
 h1, h2, h3, h4, h5, h6 {
     font-family: "Georgia", serif;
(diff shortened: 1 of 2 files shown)


r12a:
It's certainly colourful.  There are however a number of things i think we need to consider changing. Perhaps a topic for our next 1:1.  Note also the comments on a previous commit about line lengths, that were shortened by the move to a new base style sheet.

Also, note that the headings overlap the expanded TOC - another result i think of the style sheet change.

xfq:
> the headings overlap the expanded TOC

I fixed a few issues, but not this one. I couldn't find the source of this issue, and I didn't modify any z-index related style...

xfq:
> > the headings overlap the expanded TOC
> 
> I fixed a few issues, but not this one. I couldn't find the source of this issue, and I didn't modify any z-index related style...

I just found that even if I deleted all the code I added, the problem still occurs:

Screenshot 2025-06-03 at 15 08 57

------

I will take a closer look to find out what caused it.

xfq:
Should be fixed now. Please take a look.

xfq:
@r12a After I discussed with you, I further modified the style, please take a look.

r12a:
There's still a problem with overlapping content when the browser window is narrow:

Screenshot 2025-06-06 at 06 04 39

I think the width needs to be applied to the column, perhaps, rather than the menu?  (Btw, fwiw, i thought 15rem was probably wide enough when i did some testing — that also helps for narrow windows.)

xfq:
Merging.

PR #669, opened 2025-05-24
Fix #653

### LINKED ISSUE #653: [articles/missing-char-glyph/index]  Charmod reference
[source] (https://www.w3.org/International/articles/missing-char-glyph/index.en.html) [en]

> Web technology is based on the character repertoire of [Unicode](https://www.unicode.org/standard/standard.html) (see [Character Model](https://www.w3.org/TR/charmod/#sec-RefProcModel)).

Instead of charmod, maybe we should point to https://www.w3.org/International/articles/definitions-characters/index.en.html , which is more for content authors.

xfq:
[Preview](https://deploy-preview-669--i18n-drafts.netlify.app/articles/missing-char-glyph/index.en.html)

### DIFF
--- a/articles/missing-char-glyph/index.en.html
+++ b/articles/missing-char-glyph/index.en.html
@@ -57,7 +57,7 @@ f.additionalLinks = ''
     
 <section id="Problem">
 <h2>Problem description</h2>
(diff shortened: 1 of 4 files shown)


xfq:

PR #675, opened 2025-06-06
Use "characters" instead of "letters"

### LINKED ISSUE #524: [questions/qa-ltr-scripts-in-rtl]  Mirroring characters
https://www.w3.org/International/questions/qa-ltr-scripts-in-rtl#styling_mirroring

> If the font automatically reverses the images of the **letters**, that should take care of any necessary mirroring. However, the CSS transformation hack probably won't help, because it is only intended for use with short inline phrases. That said, it is perhaps less likely anyway that reversed **letters** are needed for scenarios that use the styled approach.

What needs to be mirrored is not necessarily letters, but also numbers, punctuation, word separators, etc., so I think "characters" or "character glyphs" is more accurate than "letters" here.

xfq:
Fix #524.

[Preview](https://deploy-preview-675--i18n-drafts.netlify.app/questions/qa-ltr-scripts-in-rtl.en.html#styling_mirroring)

### DIFF
--- a/questions/qa-ltr-scripts-in-rtl.en.html
+++ b/questions/qa-ltr-scripts-in-rtl.en.html
@@ -392,7 +392,7 @@ f.additionalLinks = ''
 <section id="styling_mirroring">
 <h3>Mirroring characters</h3>
 
-<p>If the font automatically reverses the images of the letters, that should take care of any necessary mirroring. However, the CSS transformation hack probably won't help, because it is only intended for use with short inline phrases. That said, it is perhaps less likely anyway that reversed letters are needed for scenarios that use the styled approach.</p>
+<p>If the font automatically reverses the images of the characters, that should take care of any necessary mirroring. However, the CSS transformation hack probably won't help, because it is only intended for use with short inline phrases. That said, it is perhaps less likely anyway that reversed characters are needed for scenarios that use the styled approach.</p>
 </section>
 </section>
 


questions/qa-ltr-scripts-in-rtl.en.html
```diff
@@ -392,7 +392,7 @@ <h2 lang="zh-hant" class="rlo notoc"><bdi class="lro">8.6%</bdi>受訪者沒頭
 <section id="styling_mirroring">
 <h3>Mirroring characters</h3>
 
-<p>If the font automatically reverses the images of the letters, that should take care of any necessary mirroring. However, the CSS transformation hack probably won't help, because it is only intended for use with short inline phrases. That said, it is perhaps less likely anyway that reversed letters are needed for scenarios that use the styled approach.</p>
+<p>If the font automatically reverses the images of the characters, that should take care of any necessary mirroring. However, the CSS transformation hack probably won't help, because it is only intended for use with short inline phrases. That said, it is perhaps less likely anyway that reversed characters are needed for scenarios that use the styled approach.</p>
```
aphillips, line 395: "it is perhaps less likely anyway" seems pretty wordy. Perhaps just "it is less likely"?

xfq: @r12a what do you think?

xfq: Because this is not directly related to the changes in this PR, I plan to merge this PR. Addison, if you think this needs to be changed, please submit a new PR. Thank you!

Issue #401, opened 2022-06-24
[questions/qa-scripts]  RTL rendering of LTR scripts


xfq:
[source] (https://www.w3.org/International/questions/qa-scripts) [en]

Should we briefly mention [RTL rendering of LTR scripts](https://w3c.github.io/i18n-drafts/questions/qa-ltr-scripts-in-rtl.en.html) in this article? Some people might take this as a special case of right-to-left scripts.

himorin:
I believe all of mentioned use cases are only historical ones, incl. Taiwanise papers, and no need to include as list of currently used.
When I've asked for some examples in Taiwanise newspaper, one at Taipei pointed that RTL was used in some political part of papers at old age but just historical for now. (so, quite hard to find live example now.)

xfq:
It is still common (although less and less) on Chinese plaques.

However, I don't think that means Chinese is an RTL script, it's just that some people might think so, so we might need to mention it in this article.

r12a:
See the PR just above.

PR #676, opened 2025-06-06
Point to RTL rendering of LTR scripts

r12a:
[Preview](https://deploy-preview-676--i18n-drafts.netlify.app/questions/qa-scripts.en.html#historical)

Relates to https://github.com/w3c/i18n-drafts/issues/401


xfq:
Looks ok, could you update the translations as well?

PR #677, opened 2025-06-06
Mention the rendering of punctuation marks

### LINKED ISSUE #558: [articles/lang-bidi-use-cases/index]  Rendering of punctuation marks
[source] (https://www.w3.org/International/articles/lang-bidi-use-cases/) [en]

I think it would be useful to add a new use case: the rendering of punctuation marks (although I'm not sure which section is the best for it).

As an example, many Chinese punctuation marks and Western punctuation marks share code points, but if the text is not labelled as Chinese, the glyphs of punctuation marks will not match the requirements for Chinese text layout.

For example:

1. `U+2018 LEFT SINGLE QUOTATION MARK [‘]`, `U+2019 RIGHT SINGLE QUOTATION MARK [’]`, `U+201C LEFT DOUBLE QUOTATION MARK [“]`, and `U+201D RIGHT DOUBLE QUOTATION MARK [”]` are non–full-width in Latin fonts, but are full-width in Chinese fonts.

2. `U+2E3A TWO-EM DASH[⸺]`, `U+2014 EM DASH [—]`, `U+2026 HORIZONTAL ELLIPSIS […]`, and `U+2013 EN DASH [–]` should be centered in the character frame, but they are not in most Latin fonts.

3. `U+00B7 MIDDLE DOT [·]` [should have the same width or half the width of a Han character](https://www.w3.org/TR/clreq/#id86), but they are not in most Latin fonts.

xfq:
Fix #558.

I'll update the translations if it's approved.

[Preview](https://deploy-preview-677--i18n-drafts.netlify.app/articles/lang-bidi-use-cases/index.en.html)

### DIFF
--- a/articles/lang-bidi-use-cases/index.en.html
+++ b/articles/lang-bidi-use-cases/index.en.html
@@ -261,6 +261,15 @@ working on.</p>
 Japanese, which are in Chinese, and which are in Tibetan. Therefore, there is a need for additional metadata to indicate the text-processing language to be used for which parts of the text, ideally with a resource-wide declaration as the default and the ability to change the language for appropriate parts of the content.</p>
 </section>
 
+<section id="punctuation_rendering">
+<h3>Punctuation mark rendering</h3>
+<p>Beyond the differences in ideographic character rendering shown in the previous section, proper language identification is also crucial for the correct rendering of punctuation marks. Many punctuation marks that are shared among scripts have different rendering requirements depending on the language context.</p>
+
+<p>For example, quotation marks such as <span class="codepoint" translate="no"><bdi lang="en">&#x2018;</bdi><code class="uname">U+2018 LEFT SINGLE QUOTATION MARK</code></span>, <span class="codepoint" translate="no"><bdi lang="en">&#x2019;</bdi><code class="uname">U+2019 RIGHT SINGLE QUOTATION MARK</code></span>, <span class="codepoint" translate="no"><bdi lang="en">&#x201C;</bdi><code class="uname">U+201C LEFT DOUBLE QUOTATION MARK</code></span>, and <span class="codepoint" translate="no"><bdi lang="en">&#x201D;</bdi><code class="uname">U+201D RIGHT DOUBLE QUOTATION MARK</code></span> appear as non-full-width characters in Latin fonts, but should be rendered as <a href="https://www.w3.org/TR/i18n-glossary/#dfn-fullwidth">full-width</a> characters in Chinese fonts.</p>
+
+<p>Without proper language tagging, browsers and other text rendering systems cannot determine which font to apply, potentially resulting in poor readability and aesthetically inconsistent text presentation.</p>
+</section>
+
 
 
 


articles/lang-bidi-use-cases/index.en.html
```diff
@@ -261,6 +261,15 @@ <h3>Capturing the text-processing language</h3>
...
 </section>
 
+<section id="punctuation_rendering">
+<h3>Punctuation mark rendering</h3>
+<p>Beyond the differences in ideographic character rendering shown in the previous section, proper language identification is also crucial for the correct rendering of punctuation marks. Many punctuation marks that are shared among scripts have different rendering requirements depending on the language context.</p>
+
+<p>For example, quotation marks such as <span class="codepoint" translate="no"><bdi lang="en">&#x2018;</bdi><code class="uname">U+2018 LEFT SINGLE QUOTATION MARK</code></span>, <span class="codepoint" translate="no"><bdi lang="en">&#x2019;</bdi><code class="uname">U+2019 RIGHT SINGLE QUOTATION MARK</code></span>, <span class="codepoint" translate="no"><bdi lang="en">&#x201C;</bdi><code class="uname">U+201C LEFT DOUBLE QUOTATION MARK</code></span>, and <span class="codepoint" translate="no"><bdi lang="en">&#x201D;</bdi><code class="uname">U+201D RIGHT DOUBLE QUOTATION MARK</code></span> appear as non-full-width characters in Latin fonts, but should be rendered as <a href="https://www.w3.org/TR/i18n-glossary/#dfn-fullwidth">full-width</a> characters in Chinese fonts.</p>
```
aphillips, line 268: This should probably be illustrated.

I would probably say "proportional" instead of "non-full-width". The characters don't necessarily change to full width so much as they space and position according to the grid. I would also probably avoid bringing font selection back a second time.

On a larger level, consider the examples I have in the UTW presentation I made:

https://www.youtube.com/watch?v=eJEKbUx3B0w&list=PLMc927ywQmTPLAnkm8qIi2yWNMZrKQdt4&index=13

Some of which are captured here:
https://w3c.github.io/i18n-discuss/explainers/lang-example.html

r12a:
I think i would have not made this a separate section. Instead, i would merge the first 2 new paragraphs into one, and insert that new para before the last paragraph in the _Capturing the text-processing language_ section. This makes it become a further example of the need for language declarations, in addition to the ones mentioned.

The 3rd new para is a useful one, which i would add at the end of the _Capturing the text-processing language_ section.

I also wondered whether it might be useful to add this example to https://www.w3.org/International/questions/qa-lang-why.en.html

articles/lang-bidi-use-cases/index.en.html
xfq: > This should probably be illustrated.

OK.

> I would probably say "proportional" instead of "non-full-width".

Fine with me. I am used to using "non-full-width" because I often refer to both "proportional" and "half-width" (glyphs that take up exactly half the width of a Han character) at the same time, but "proportional" can indeed be used here.

> 
> On a larger level, consider the examples I have in the UTW presentation I made:
> 
> https://www.youtube.com/watch?v=eJEKbUx3B0w&list=PLMc927ywQmTPLAnkm8qIi2yWNMZrKQdt4&index=13
> 
> Some of which are captured here: https://w3c.github.io/i18n-discuss/explainers/lang-example.html

This example has already been mentioned in the article. It is an example of CJK shared code points. What I want to explain here is the example of Latin and non-Latin shared code points.

PR #681, opened 2025-06-09
Add a JavaScript Internationalization article

### LINKED ISSUE #659: JavaScript Internationalization (Intl API)
Modern web apps rely on client‐side code for localization (e.g. formatting dates, numbers, and plural forms). An overview and some examples on using the ECMAScript Internationalization API (Intl.DateTimeFormat, Intl.NumberFormat, Intl.Collator, etc.) would be useful.

xfq:
Fix #659.

I think it is very useful for website authors/developers, and IMHO the content of this PR as an overview is easier to understand than the content on the Ecma spec and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl).

We need to decide where to put this in https://www.w3.org/International/articlelist . Should we add a new scripting section, or should we add it to one of the existing sections?

[Preview](https://deploy-preview-681--i18n-drafts.netlify.app/articles/intl/index.en.html)


xfq:
Merging per https://github.com/w3c/i18n-actions/issues/183

PR #693, opened 2025-06-19
Initial draft on number, currency, and unit formatting

### LINKED ISSUE #410: An article about number, currency, and unit formatting
Can we write an article about currency? It's useful for both spec developers and content authors.

Here are some concepts we can mention:

## Decimal separator

Some regions use a period (.) as a separator to separate the integer part from the fractional part. Some regions use a comma (,). Some regioins use the currency symbol (such as the [Cape Verdean escudo](https://en.wikipedia.org/wiki/Cape_Verdean_escudo)).

Currencies can have two decimals (like GBP), three decimals (like BHD), or no decimals (like VND).

## Digit grouping

Some regions use a comma to separate thousands. Some regions use a period for this purpose.

Some regions use 2-digit grouping or 4-digit grouping.

## Negative currency

The negative sign can be used:

* before both the currency symbol and number
* before the number but after the currency symbol
* after the number

Alternatively, the number can be displayed with parentheses around it or even in a colour such as red.

## Numbering system

Numbers can be [European numerals](https://w3c.github.io/clreq/#term.european-numerals) or locale-specific (such as Chinese/Japanese or Devanagari).

## Currency symbol

It can be a pre-defined symbol such as the European Euro '€' or a combination of letters like the use of 'CNY' for Renminbi yuan.

It can be placed before or after the digits.

The same currency symbol may represent multiple currencies.

## API

When designing a currency formatting API, it should:

* let the content author choose whether to show the digit grouping (like the thousands separator)
* let the content author round the number, and choose among `ceil`, `floor`, and `round`
* support symbols like K, M, B, T
* enable content authors to customize their own currency formats such as `$123.4万`

xfq:
Fix #410. Comments welcome.

[Preview](https://deploy-preview-693--i18n-drafts.netlify.app/questions/qa-number-format.en.html)


questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+### The core problem: why is formatting so complex?
+
+Before diving into solutions, let's unpack the nuances that make international number and currency formatting a non-trivial task:
+
+#### Decimal separators
+
+Many regions, like the US and UK, use a period as the decimal separator (e.g., `1,234.56`). Conversely, much of Europe and South America use a comma (e.g., `1.234,56`).
+
+In rare cases, a currency symbol can even act as a decimal separator, like the [Cape Verdean escudo](https://en.wikipedia.org/wiki/Cape_Verdean_escudo).
```
r12a: Add an example here.

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Numbering systems
+
+While [European digits](https://www.w3.org/TR/i18n-glossary/#dfn-european-digits) (0-9) are globally prevalent, many cultures prefer their native numeral systems. For instance, 12,345 might be `١٢٬٣٤٥` in Arabic or `๑๒,๓๔๕` in Thai.
+
+#### Currency symbol display
+
+**Symbol vs. Code vs. Name:** Currencies can be displayed using a specific symbol (e.g., `€`, `£`, `¥`), their ISO 4217 code (e.g., `EUR`, `GBP`, `JPY`), or their full name (e.g., `US Dollar`, `Japanese Yen`).
+
+**Placement:** The symbol can appear before the number (e.g., `$100.00`) or after (e.g., `1,000 ₫`).
```
r12a: The example illustrates that white space also varies.

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+    unitDisplay: 'long',
+});
+console.log(`French Kilobytes: ${kbFrFormatter.format(storage)}`); // Output: French Kilobytes: 5 kilooctets
+```
+
+You can find a list of standard unit identifiers (e.g., `meter`, `kilogram`, `liter`, `kilobyte`, `percent`, `hour`) in the [ECMAScript Internationalization API Specification](https://tc39.es/ecma402/#sec-issanctionedsimpleunitidentifier).
+
+## By the way
+
+Typically, you'll want to format based on the user's browser language. You can get this from `navigator.language` or `navigator.languages`:
```
r12a: Actually, this probably merits its own section.  There may be times when it's best to use the browser settings, but usually its the page language or current section language that is more important.

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+```javascript
+const userLocale = navigator.language || 'en-US'; // Fallback to 'en-US'
+const formatter = new Intl.NumberFormat(userLocale, { style: 'currency', currency: 'EUR' });
+console.log(`Price for user's locale: ${formatter.format(199.99)}`);
+```
+
+## Further reading
+
+* [ECMAScript Internationalization API Specification](https://tc39.es/ecma402/)
```
r12a: We should add some dos and donts to [the Web](https://www.w3.org/International/techniques/authoring-html) and point to that, too.

questions/qa-number-format.en.md
aphillips: I was going to thumbs-up this comment, but I'd be more emphatic. _Typically_ you want to use the locale (language) of the page. _Occasionally_ you'll want to override that with an author specified locale (such as when writing I18N tutorials). Rarely will you want to peek at browser language (and there is no way to do so at the server--`Accept-Language` isn't quite the same thing).

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
+# Number, currency, and unit formatting
+
+## Question
+
+**How do I prepare my web pages to display varying international number, currency, and unit formats?**
```
aphillips: I'm not sure this is the right question for what this article proceeds to talk about? JS is a mechanism for formatting data into strings dynamically in the page. Some parts of a Web page will need to do that, but this question sounds like it would cover server-side formatting.

---

I'll also call out that numbers and currencies are a broad topic. Units are complicated and might warrant separate coverage. (Currencies are complicated too, but are generally bundled with numbers. Don't overlook percentages, which are also commonly bundled, but which are a sort of unit.)

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+Number, currency, and unit formatting varies dramatically across cultures and regions. A seemingly innocuous error can sometimes lead to financial misinterpretations.
```
aphillips: I would says "... cultures, regions, and languages." I would also use the term locale somewhere near the start---for our programmer friends 😄 

"financial misinterpretations" only pertains to currencies.

```suggestion
The formats used by numbers, including specialized formatting such as currencies and units,
varies dramatically across cultures, regions, and languages.
A seemingly innocuous error can lead misunderstanding or mistakes.
```

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Answer
+
+Hardcoding formats is a brittle and unsustainable approach. Fortunately, the web platform provides robust, standards-based solutions to navigate this complex landscape. This article will guide you, step-by-step, through leveraging JavaScript's built-in `Intl` object to effortlessly adapt your web pages to varying international number, currency, and unit formats.
+
+### The core problem: why is formatting so complex?
+
+Before diving into solutions, let's unpack the nuances that make international number and currency formatting a non-trivial task:
```
aphillips: You lost units here 😉 so maybe lose it elsewhere...

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+Grouping patterns also differ. **3-digit grouping** (e.g., `1,234,567`) is common, but some regions, like India, use a **2-digit grouping** beyond the hundreds (e.g., `12,34,567`). Some specialized formats might use 4-digit grouping.
+
+#### Numbering systems
+
+While [European digits](https://www.w3.org/TR/i18n-glossary/#dfn-european-digits) (0-9) are globally prevalent, many cultures prefer their native numeral systems. For instance, 12,345 might be `١٢٬٣٤٥` in Arabic or `๑๒,๓๔๕` in Thai.
+
+#### Currency symbol display
+
+**Symbol vs. Code vs. Name:** Currencies can be displayed using a specific symbol (e.g., `€`, `£`, `¥`), their ISO 4217 code (e.g., `EUR`, `GBP`, `JPY`), or their full name (e.g., `US Dollar`, `Japanese Yen`).
```
aphillips: This is not the full range. There are also "medium" width symbols used in various contexts, such as `CA$`, `US$` and `MX$`

There are also contextual variations, e.g. `¥` vs. `円` (in sentences).

Symbols are not always a single code point and you should probably have some examples of those.

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Placement:** The symbol can appear before the number (e.g., `$100.00`) or after (e.g., `1,000 ₫`).
+
+**Ambiguity:** The same symbol might represent multiple currencies (e.g., `$` for US Dollar, Canadian Dollar, Australian Dollar, etc.). So the locale is crucial.
```
aphillips: The locale is crucial all the time. Note that the Mexican peso also uses `$` and makes a good example (showing that `$` doesn't mean the word _dollar_ is lurking nearby)

aphillips (commented):
Partial review.

questions/qa-number-format.en.md
xfq: > I'm not sure this is the right question for what this article proceeds to talk about? JS is a mechanism for formatting data into strings dynamically in the page. Some parts of a Web page will need to do that, but this question sounds like it would cover server-side formatting.

What about: **How do I use JavaScript to dynamically format numbers, currencies, and units for different locales?**

-----

> I'll also call out that numbers and currencies are a broad topic. Units are complicated and might warrant separate coverage. (Currencies are complicated too, but are generally bundled with numbers. Don't overlook percentages, which are also commonly bundled, but which are a sort of unit.)

Any specific suggestions? Should we expand this article or start a new one?

questions/qa-number-format.en.md
xfq: Fixed. Thanks.

questions/qa-number-format.en.md
xfq: fixed

questions/qa-number-format.en.md
xfq: The double barrd dollarsign doesn't seem to be encoded in Unicode. Should I use a picture to represent it?

questions/qa-number-format.en.md
xfq: Fair enough. I expanded this paragraph.

questions/qa-number-format.en.md
xfq: Added.

questions/qa-number-format.en.md
xfq: Good point. Fixed.

questions/qa-number-format.en.md
xfq: Makes sense. I edited the content of this section.

questions/qa-number-format.en.md
xfq: Yes, I'm going to add the dos and donts after this article is published so that I can link to it.

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,194 @@
...
+- **ISO 4217 codes:** Three-letter standardized codes (e.g., `EUR`, `GBP`, `JPY`)
+- **Full names:** Complete currency names (e.g., `US Dollar`, `Japanese Yen`)
+
+**Placement and spacing:** The symbol can appear before the number (e.g., `$100.00`) or after (e.g., `1,000 ₫`). From these two examples, we can also see that some currencies use no space between the symbol and number, while others include a space.
+
+**Ambiguity:** The same symbol might represent multiple currencies (e.g., `$` for US Dollar, Canadian Dollar, Mexican Peso, etc.).
+
+### The `Intl` object
+
+The `Intl` object in JavaScript automatically handles locale-specific decimal and grouping separators, currency symbols, and other numerical conventions.
```
jsahleen: Should this be the `Intl.NumberFormat` object? Or maybe `the JavaScript Intl API`?

questions/qa-number-format.en.md
```diff
@@ -0,0 +1,194 @@
...
+    unitDisplay: 'long',
+});
+console.log(`French Kilobytes: ${kbFrFormatter.format(storage)}`); // Output: French Kilobytes: 5 kilooctets
+```
+
+You can find a list of standard unit identifiers (e.g., `meter`, `kilogram`, `liter`, `kilobyte`, `percent`, `hour`) in the [ECMAScript Internationalization API Specification](https://tc39.es/ecma402/#sec-issanctionedsimpleunitidentifier).
+
+### Choosing the right locale
+
+Typically, you'll want to format number, currency, and unit according to the language of the web page. This can be determined from the `lang` attribute:
```
jsahleen: I think something about the wording here might lead people to think the `lang` attribute is populated by default, when it needs to be set. This isn't the right place for an in-depth discussion, I think, but is there an article or guide on choosing the right locale we can point to?

jsahleen (commented):
Great article! Mentioned a couple of minor points but nothing blocking.

questions/qa-number-format.en.md
jsahleen: Maybe this: https://www.w3.org/International/questions/qa-html-language-declarations.en.html

r12a: @jsahleen if you want an article that's specifically about choosing the right locale you may be looking for https://www.w3.org/International/questions/qa-choosing-language-tags.en.html

Otherwise, there are a bunch of other language setting related articles at https://www.w3.org/International/articlelist#language

questions/qa-number-format.en.md
xfq: fixed, thanks

questions/qa-number-format.en.md
xfq: Thanks. I added a link.

jsahleen (approved):

questions/qa-number-format.en.md
```diff
@@ -46,7 +46,7 @@ While [European digits](https://www.w3.org/TR/i18n-glossary/#dfn-european-digits
 
 ### The `Intl` object
```
jsahleen: Does this heading also need to be changed?

xfq: fixed, thanks!

PR #705, opened 2025-07-01
Use "text content" instead of "content"

### LINKED ISSUE #691: [tutorials/tutorial-char-enc/index]  Content vs text content
[source] (https://www.w3.org/International/tutorials/tutorial-char-enc/index) [en]

> Content is composed of a sequence of characters. Characters represent letters of the alphabet, punctuation, etc. But content is stored in a computer as a sequence of bytes, which are numeric values.

I think it’s better to say "**Text** content is composed of a sequence of characters", because a web page can have other non-text content such as images, which are not composed of characters.

xfq:
Fix #691.


xfq:

PR #726, opened 2025-07-24
Update qa-personal-names.en.html re capitalisation

HedvigS:
I encountered a situation with French people where they like to write their family/last/surname in all upper case and thought it might be worth mentioning. Feel free to ignore if overkill.

### DIFF
--- a/questions/qa-personal-names.en.html
+++ b/questions/qa-personal-names.en.html
@@ -235,7 +235,7 @@ f.additionalLinks = ''
           <input id="fullname3" value="" type="text" style="width:250px;" />
         </p>
         <p>
-          <label for="salutation">What  should we call you? <span style="color:#888;">(for example, when we send you mail?)</span></label>
+          <label for="salutation">What  should we call you? <span style="color:#888;">(for example, when we send you mail)</span></label>
           <br>
           <input id="salutation" value="" type="text" style="width:250px;" />
         </p>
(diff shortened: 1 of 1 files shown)


questions/qa-personal-names.en.html
```diff
@@ -268,7 +268,7 @@ <h3>Other things</h3>
...
-<p><mark id="req_case_normalization">Don't normalize the casing in names.</mark>Some names (such as 'McNamara') contain capital letters that are not the first letter; others (such as 'van der Waals') include words that are not capitalized.  Forms should preserve the case the user enters and not coerce such names to always and only use capital letters at the start of each word. </p>
+<p><mark id="req_case_normalization">Don't normalize the casing in names.</mark>Some names (such as 'McNamara') contain capital letters that are not the first letter; others (such as 'van der Waals') include words that are not capitalized.  Forms should preserve the case the user enters and not coerce such names to always and only use capital letters at the start of each word. Note, in France many people write their family/last/surname all in upper case (e.g. Paul SIMON) in order to distinguish it from first/given names with the same spelling. It can be tempting to therefore normalise all casing, but this is not desirable as it would cause issues in other instances, such as "van der Meer" or "McDonald". </p>
```
r12a, line 271: It's not only in France that this uppercasing occurs.  I'm also not sure whether it occurs much in form input...

HedvigS: Yes, France is just one example.
I've seen it in form input from people from Asia too.

PR #764, opened 2025-08-22
Add a visual example of hyphenation

xfq:
I'll update the translations if this is approved.

[Preview](https://deploy-preview-764--i18n-drafts.netlify.app/questions/qa-lang-why.en.html#rendering)


r12a:
The figure looks ok, but i think it would be useful to point to a test from the figure.  I created such a test at https://github.com/w3c/line_paragraph_tests/issues/96

r12a (approved):

PR #769, opened 2025-09-01
Try adding mobile CSS

### LINKED ISSUE #299: Make the layout look better on mobile browsers
The pages don't seem to be mobile optimized at the moment. It would be great if we can make them look better on narrow screen devices in the future.

xfq:
Fix #299.

[Preview](https://deploy-preview-769--i18n-drafts.netlify.app/nav/about.html)


r12a:
Looks good overall.  I spotted one issue though.  The What is "I18n"? aside needs its left margin set flush with the left side of the page, and there should be more space between the text of the aside and the following H1 heading – again, probably easiest to achieve by increasing the margin-block-end spacing for the aside.

Screenshot 2025-09-04 at 11 13 11

r12a:
The latest changes are an improvement, but it still looks a bit clunky at the top of the page. How about we reduce the font-size for the two headings that would normally be in the right column?  When i reduced to 100% it looks like this:

Screenshot 2025-10-09 at 11 07 40

xfq:
> The latest changes are an improvement, but it still looks a bit clunky at the top of the page. How about we reduce the font-size for the two headings that would normally be in the right column?

Thanks for the suggestion, I adjusted the font size.

PR #783, opened 2025-10-21
Try updating the screenshot of Unicode

xfq:
From 5.2 to 17.

If it's approved, I'll optimize the PNG file and update the translations.

[Preview](https://deploy-preview-783--i18n-drafts.netlify.app/articles/definitions-characters/index.en.html)


r12a:
There seems to be a lack of white space above and, mostly, below the image.  I think we should also change the introductory paragraph to say that the image shows just a small number of Unicode scripts.  The aim of the previous image was to show just how many scripts and other blocks are supported by Unicode.

xfq:

PR #792, opened 2025-11-12
Update UTF-8 data

xfq:


xfq:
Merging. Translations updated as well.

Issue #521, opened 2023-08-26
URL structure for internationalized websites


xfq:
It might be useful to write an article about how to design a URL structure for an internationalized website. We can explain the pros and cons of each approach and let the content authors judge for themselves. There's some relevant stuff in [Monolingual vs. multilingual Web sites](https://www.w3.org/International/questions/qa-mono-multilingual.en.html), but it didn't mention URLs.

Some of the common URL structures for internationalized websites include:

## Different domains

In rare cases, companies use entirely different domains, like `douyin.com` in China and `tiktok.com` for the rest of the world. This is more of a branding/legal strategy than a typical i18n approach.

## ccTLDs (including IDN ccTLDs)

This approach involves using a separate, country-specific domain for each target market.

Examples:

* google.co.jp
* apple.com.cn
* amazon.co.jp
* philips.com.hk

Using this approach requires spending money to buy register new domain names, but you can use different servers for different regions.

## Subdomains

This approach uses a different third-level domain for each language or region.

Examples:

* zh.wikipedia.org
* hk.yahoo.com
* uk.norton.com

It's technically straightforward to create new subdomains.

## Folders

Create a folder within the main site for each language or locale.

Examples:

* https://www.samsung.com/hk/
* https://www.apple.com/jp/
* https://support.zoom.us/hc/en-us
* https://www.adobe.com/hk_en/

## Query string

This method appends the language/locale to the URL using a `?` parameter.

Examples:

* https://www.office.com/?omkt=zh-CN
* https://developers.google.com/?hl=zh-cn
* https://www.linkedin.com/help/linkedin?lang=ro-RO

## Change the file name

Example:

https://www.w3.org/International/questions/qa-site-conneg.sv.html

## No change to the URL

Examples:

* Facebook (main site, the support pages use query strings)
* LinkedIn (main site, the support pages use query strings)
* https://web.dev/

aphillips:
This might be a useful article, although please note that the various URL elements you cite are doing different things. In some cases they convey locale variation. In other cases they refer to region-specific sites or region-specific content. Language negotiation and preference management might be a good subset: I seem to have written that down multiple times of late...

himorin:
For change the file name one, this was a commonly used multiple file extensions' option, which was a part of mod_mime.c. And there was no defined order among multiple file extensions, so both `.<lang>.<ext>` and .`<ext>.<lang>` were handled as the same. (in some special case, when a file extension is tighten to specific data handler, order has some mean, although...)
[...]

PR #795, opened 2025-11-14
Add information about search engine crawling and indexing

### LINKED ISSUE #794: [questions/qa-when-lang-neg]  Challenges for search engine crawling and indexing
[source] (https://www.w3.org/International/questions/qa-when-lang-neg.en.html) [en]

One of the primary issues with language negotiation is how search engines discover and index web content. Search engine crawlers do not behave like typical users. This discrepancy creates challenges for websites that rely on the `Accept-Language` header to serve different language versions of a page under the same URL.

A major hurdle is that search engine crawlers do not consistently send an `Accept-Language` header when requesting a page. Googlebot, for instance, [often crawls without this header](https://developers.google.com/search/docs/specialty/international/locale-adaptive-pages), or may default to English. This means the crawler may only ever see the default language version of a page. Consequently, other language versions of the content may not be discovered, crawled, or indexed, making them invisible to users searching in those languages.

Also, presenting different content to search engines and users is a practice known as "cloaking", which can be penalized by search engines. While language negotiation is not inherently cloaking, if not implemented carefully, it can be misconstrued as such.

To circumvent these issues, search engines like Google explicitly recommends using separate URLs for each language version of a page. This approach, combined with the use of `hreflang` annotations, provides a clear signal to search engines about the different language variations of a page and their relationship to one another. `hreflang` helps search engines understand which language or regional version of a page to show to a user based on their language and location settings.

See also https://github.com/w3c/i18n-drafts/issues/666 and https://github.com/w3c/i18n-drafts/issues/405

xfq:
Fix #794.

[Preview](https://deploy-preview-795--i18n-drafts.netlify.app/questions/qa-when-lang-neg.en.html)

### DIFF
--- a/questions/qa-when-lang-neg.en.html
+++ b/questions/qa-when-lang-neg.en.html
@@ -20,7 +20,7 @@ f.status = 'published'  // should be one of draft, review, published, notreviewe
 f.path = '../' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2022-05-30', time:'16:32'} // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2025-11-14', time:'14:26' } // date and time of latest edits to this document/translation
 f.contributors = '' // people providing useful contributions or feedback during review or at other times
 // make sure to translate the information in the meta description element
 // also make sure that the lang attribute on the html tag is correct!
@@ -74,10 +74,10 @@ f.additionalLinks = ''
 
 <section id="answer">
 <h2>Answer</h2>
-  <p>The short answer is: <em>always</em>.</p>
-  <p>A slightly longer answer is: <em>almost</em> always, <em>but not alone</em>.</p>
-  <p>Language negotiation <a href="#shortcomings">does not always work as intended</a>; there are techniques to <a href="#overrides">make up
-    for the shortcomings</a>; one should also provide for <a href="#stickyness">stickiness in navigation</a>.</p>
+  <p>The short answer is: <em>almost always, but not alone</em>.</p>
+  <p>Language negotiation is a useful feature that improves user experience by serving content in preferred languages. However, it has significant limitations that must be addressed.</p>
+  <p>Language negotiation <a href="#shortcomings">does not always work as intended</a>. There are techniques to <a href="#overrides">make up
+    for the shortcomings</a>, and one should also provide for <a href="#stickyness">stickiness in navigation</a>.</p>
     
   <section id="shortcomings">
 <h3>Shortcomings of language negotiation</h3>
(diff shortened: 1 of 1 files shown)


xfq:
After discussing with @r12a, we decided to limit this PR to the "When to use language negotiation" part and put "How to do it" (such as the last two paragraphs about `hreflang` and language-specific URLs) in a new article.

The new article should probably be in [Navigation](https://www.w3.org/International/articlelist#navigating). Also related to #521.

I'll try to do this.

questions/qa-when-lang-neg.en.html
```diff
@@ -101,6 +101,14 @@ <h3>Shortcomings of language negotiation</h3>
       sub-sections. Note however that some measure of cultural adaptation (e.g. changing the currency) does not necessarily make pages non-equivalent; the
       non-equivalence limitation to language negotiation really exists when a <em>site</em> is adapted so that pages in different languages do not
       correspond to one another any more.</p>
+
+    <section id="searchengines">
+    <h4>Search engine discovery challenges</h4>
+      <p>A major limitation of language negotiation is how search engines discover and index multilingual content. Search engine crawlers often do not send an <code class="kw" translate="no">Accept-Language</code> header when requesting pages, or may default to a specific language such as English. For example, <a href="https://developers.google.com/search/docs/specialty/international/locale-adaptive-pages" target="_blank">Googlebot frequently crawls without this header</a>. This means the crawler may only discover and index the default language version of a page. Consequently, other language versions remain invisible to users searching in those languages, significantly limiting the reach of multilingual content.</p>
```
jsahleen, line 107: Suggest: "Another limitation of language negotiation has do with the way search engines discover and index multilingual content."

questions/qa-when-lang-neg.en.html
```diff
@@ -101,6 +101,14 @@ <h3>Shortcomings of language negotiation</h3>
...
+      <p>To address these issues, search engines explicitly recommend using separate URLs for each language version of a page. This approach provides clear signals to search engines about available language variations. Combined with <code class="kw" translate="no">hreflang</code> annotations in the HTML <code class="kw" translate="no">head</code> or in XML sitemaps, separate URLs help search engines understand which language or regional version to show users based on their language and location settings.</p>
+      <p>If you implement language negotiation, you should also provide language-specific URLs (such as example.com/de/page.html for German and example.com/fr/page.html for French) that can be discovered and indexed by search engines. The language-generic URL (example.com/page.html) can still use content negotiation for direct visitor access, but the language-specific URLs ensure all versions are discoverable.</p>
```
jsahleen: This only applies if you want the pages to be discoverable and indexable, right?

jsahleen (approved):
A couple suggestions/questions, but nothing that should block merging.

questions/qa-when-lang-neg.en.html
xfq: Yes, as I said [above](https://github.com/w3c/i18n-drafts/pull/795#issuecomment-3558041930), I plan to remove this paragraph and the previous paragraph and move them into a new article.

jsahleen: Ah. Yes. Apologies. I wasn't clear on the intent.

xfq: Moved to https://github.com/w3c/i18n-drafts/pull/799

questions/qa-when-lang-neg.en.html
```diff
@@ -74,10 +74,10 @@ <h2>Question</h2>
 
 <section id="answer">
 <h2>Answer</h2>
```
aphillips, line 76: These changes are unrelated to the title. You should pull them separately?

questions/qa-when-lang-neg.en.html
```diff
@@ -101,6 +101,12 @@ <h3>Shortcomings of language negotiation</h3>
       sub-sections. Note however that some measure of cultural adaptation (e.g. changing the currency) does not necessarily make pages non-equivalent; the
       non-equivalence limitation to language negotiation really exists when a <em>site</em> is adapted so that pages in different languages do not
       correspond to one another any more.</p>
+
+    <section id="searchengines">
+    <h4>Search engine discovery challenges</h4>
```
aphillips, line 106: This section makes a number of assumptions about how language negotiated pages are structured, stored, and served. Those assumptions may not hold true in practice and are an important consideration in how a given site structures its language negotiation. For example, in my proposed article about language negotiation, I call out the need for domain, path, or query based language identification. One reason for this is to expose per-page language/locale to robots and crawlers.

This also helps with things like offline links in a specific language (if I have a display ad in Spanish, I want the link to to go the Spanish language experience page)

xfq: Do you have any specific suggestions regarding the text here? After discussing with Richard, we decided to limit the content of this document to the "when" (and a bit "why") aspects, while the "what" and "how" parts will be discussed in other articles.

xfq:
Discussed in today's WG telecon: https://www.w3.org/2026/01/15-i18n-minutes.html#de51

PR #799, opened 2025-11-21
Add guide to proper language negotiation

### LINKED ISSUE #798: Guide to proper language negotiation
per https://github.com/w3c/i18n-drafts/pull/795#issuecomment-3558041930 , I'll create an article about how to set up language negotiation properly.

xfq:
Fix #798.

This is a first draft. Feedback is welcome.

[Preview](https://github.com/w3c/i18n-drafts/blob/xfq/issue-798/articles/proper-lang-neg/index.md)

### DIFF
--- a/articles/proper-lang-neg/index.md
+++ b/articles/proper-lang-neg/index.md
@@ -6,7 +6,7 @@
 
 Language negotiation is the mechanism by which a server selects the best version of a page based on the client’s `Accept-Language` header. It is a concept of the HTTP protocol. As noted in [When to use language negotiation](https://www.w3.org/International/questions/qa-when-lang-neg), the short answer to "when to use language negotiation" is *always*. However, the long answer is *always, but never alone.*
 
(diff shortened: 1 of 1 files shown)


aphillips:
Don't forget about this document hanging out in development: https://github.com/w3c/i18n-drafts/blob/gh-pages/articles/language-negotiation/language-negotiation.md

There are a few things in this doc that also belong in the other. Or maybe some things in there that belong here. But we shouldn't end up with _competing_ articles. We should have a separation of concerns if we end up with two docs.

xfq:
> Don't forget about this document hanging out in development: https://github.com/w3c/i18n-drafts/blob/gh-pages/articles/language-negotiation/language-negotiation.md
> 
> There are a few things in this doc that also belong in the other. Or maybe some things in there that belong here. But we shouldn't end up with _competing_ articles. We should have a separation of concerns if we end up with two docs.

Here's my idea: we will have three articles related to language negotiation:

- https://github.com/w3c/i18n-drafts/blob/gh-pages/articles/language-negotiation/language-negotiation.md mainly discusses "what is language negotiation";
- https://www.w3.org/International/questions/qa-when-lang-neg discusses "when language negotiation is needed";
- this PR discusses "how to do it".

All three articles include some content on "why language negotiation is needed". What do you think of this approach? If you agree, I'll think about how to adjust these articles to minimize repetition.

PR #801, opened 2025-12-12
Update job description for r12a

r12a:
We need to also update co-chairs, but i don't know whether Joel wants his email to be listed. @xfq could you take care of that?


xfq:
> We need to also update co-chairs, but i don't know whether Joel wants his email to be listed. @xfq could you take care of that?

Will do.

PR #815, opened 2026-01-15
Mention ruby specs in the Ruby Markup article

xfq:
[Preview](https://deploy-preview-815--i18n-drafts.netlify.app/articles/ruby/markup.en.html#endlinks)


xfq:
Merging per [i18n telecon](https://www.w3.org/2026/01/15-i18n-minutes.html#6336). Copied to CVS as well.

PR #834, opened 2026-03-11
Add quick list of RTL languages

xfq:
[Preview](https://deploy-preview-834--i18n-drafts.netlify.app/questions/qa-scripts.en.html)

### DIFF
--- a/questions/qa-scripts.en.html
+++ b/questions/qa-scripts.en.html
@@ -15,12 +15,12 @@ f.previousauthors = '' // as above
 f.modifiers = 'Richard Ishida, W3C' // people making substantive changes, and their affiliation
 f.searchString = 'qa-scripts' // blog search string - usually the filename without extensions
 f.firstPubDate = '2003-09-12' // date of the first publication of the document (after review)
-f.lastSubstUpdate = { date:'2026-03-11', time:'12:53'}  // date and time of latest substantive changes to this document
+f.lastSubstUpdate = { date:'2025-06-12', time:'09:40'}  // date and time of latest substantive changes to this document
 f.status = 'published'  // should be one of draft, review, published, notreviewed or obsolete
 f.path = '../' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2026-03-11', time:'12:58' } // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2026-03-11', time:'13:00' } // date and time of latest edits to this document/translation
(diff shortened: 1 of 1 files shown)


r12a:
Ah, i made a comment on the commit, but i should have waited another 3 minutes and i'd have seen this PR.

My comment was: When it refers to RTL languages, this change reintroduces wording we previously fixed. We even have a subsection that explains that languages aren't RTL, only scripts that are used for those languages.

questions/qa-scripts.en.html
```diff
@@ -128,6 +128,8 @@ <h3>Which languages are written right-to-left (RTL)?</h3>
 <section id="directions">
 <h3>What languages use RTL scripts?</h3>
 
+<p>If you need a quick list of languages commonly described as RTL languages, start with Arabic, Hebrew, Persian, Urdu, Pashto, Sindhi, Syriac, Thaana, N'Ko, and Adlam.</p>
```
r12a, line 131:
```suggestion
<p>If you need a quick list of languages that use RTL scripts, start with Arabic, Hebrew, Persian, Urdu, Pashto, Sindhi, Syriac, Thaana, N'Ko, and Adlam.</p>
```
What criteria were used to select these and not others? And to order them?

r12a (changes requested):

questions/qa-scripts.en.html
xfq: Regarding this point, I considered various approaches and I specifically chose to use this particular phrasing because a great many people use the term "RTL language" to describe this concept. Therefore, I opted for the phrase "commonly described as" in order to articulate this idea in a manner that is more familiar to a wider audience.

As for the sorting order, this text does not currently have a specific order. I can certainly arrange them in alphabetical order.

xfq: Like [this PR](https://github.com/w3c/clreq/pull/742), it describes a technically inaccurate term, yet one that is widely used by a great many people.

r12a: I think that, even if people do use that phrasing, it's not necessary to use it here to get the point across.  And since the article has a section that takes the trouble to point out that such terminology is inaccurate (eg. Azerbaijani is also a 'left-to-right language'), it introduces a strange inconsistency to then use that phrasing in the article .  Bottom line, though, is that i don't see what we gain by using that phrase here, when the alternative is simpler and more accurate.  If you want to point out that people sometimes commonly describe these languages as "RTL languages", i'd have thought that it would be better to mention that in the section "Which languages are written right-to-left (RTL)?".

xfq:
After giving it some thought, I decide to close this PR for now. I'll consider opening a new one once I come up with better wording. Thanks for your feedback, @r12a!

PR #844, opened 2026-03-19
Refine pinyin annotation description

### LINKED ISSUE #698: [questions/qa-ruby]  Placement of pinyin transliterations
[source] (https://www.w3.org/International/questions/qa-ruby.en.html) [en]

> In Mainland China, Simplified Chinese characters are typically annotated with pinyin transliterations above or below each character.

This is not very accurate. Most of the time pinyin transliterations are above each character, and it is rare to be below each character.

xfq:
Fix #698.

I'll update the translation if this is approved.


xfq:
Merging per discussions offline. Updated the translation and copied to CVS as well.

PR #849, opened 2026-03-24
Refine bidi styling guidance in qa-bidi-css-markup

### LINKED ISSUE #717: [questions/qa-bidi-css-markup]  CSS vs. markup for bidi support
[source] (https://www.w3.org/International/questions/qa-bidi-css-markup.en.html) [en]

> For HTML, the bidi markup is sufficient and no CSS bidi styling is needed.

"CSS bidi styling" is a bit vague. For example, the `:dir()` selector is still sometimes needed.

xfq:
Fix #717.

I'll update the translation if this is approved.

[Preview](https://deploy-preview-849--i18n-drafts.netlify.app/questions/qa-bidi-css-markup.en.html)

### DIFF
--- a/questions/qa-bidi-css-markup.en.html
+++ b/questions/qa-bidi-css-markup.en.html
@@ -20,7 +20,7 @@ f.status = 'published'  // should be one of draft, review, published, notreviewe
 f.path = '../' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2026-03-24', time:'16:27' } // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2026-03-31', time:'21:39' } // date and time of latest edits to this document/translation
 f.contributors = '' // people providing useful contributions or feedback during review or at other times
 // make sure to translate the information in the meta description element
 // also make sure that the lang attribute on the html tag is correct!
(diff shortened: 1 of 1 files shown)


r12a (changes requested):
This section is supposed to summarise the content below it. The statement about :dir() is not part of that content. We should either add a paragraph explaining what this does to the content below, or point to another article that explains how to use :dir().

xfq:
Thanks for the review. I adjusted it in a different way, WDYT?

questions/qa-bidi-css-markup.en.html
```diff
@@ -71,7 +71,7 @@ <h2>Question</h2>
 <section id="nutshell">
     <h2 id="quickanswer">Quick answer</h2>
     <p>You should always use dedicated bidi markup to describe your content, where markup is available.</p>
-    <p>For HTML, the bidi markup is sufficient and no CSS bidi styling is needed. You should never apply CSS bidi styling to ordinary elements or attributes. This includes content delivered as XML  using XHTML 1.x or polyglot markup.</p>
+    <p>For HTML, the bidi markup is sufficient, and you should not use CSS properties such as <code class="kw" translate="no">direction</code> or <code class="kw" translate="no">unicode-bidi</code> to control bidirectional behavior. CSS selectors such as <code class="kw" translate="no">:dir()</code> can still be useful for applying other styles that depend on the direction of text, but they do not control bidirectional behavior. That is a separate styling topic, and is not what the detailed discussion below is about.</p>
```
aphillips, line 74: I would reformulate this section differently.

---

> HTML provides comprehensive support for bidi using the `dir` attribute, as well as, sparingly, the `bdi` and `bdo` elements for special cases. Avoid tailoring the CSS properties `direction` and `unicode-bidi`.
>
> CSS selectors such as `:dir()` are useful for applying other styles that depend on the direction of text. These selectors do not control the text direction or alter how the bidi algorithm is applied to text. Instead they are used for altering presentational styles such as "left" vs. "right" position and sizing for thinks like borders, margins, alignment and the like.

aphillips (commented):
Suggesting a reformulation.

PR #856, opened 2026-03-31
Update sources in qa-scripts

### LINKED ISSUE #833: [questions/qa-scripts]  Data sourcing is too thin
[source] (https://www.w3.org/International/questions/qa-scripts) [en]

The article gives some numbers like “12 scripts”, “215 languages”, and 2,305,048,719 users, but only says the data comes from Ethnologue, without an edition or retrieval date. It makes the numbers hard to trust or maintain.

xfq:
Fix #833.

I'll update the translations if this is approved.

[Preview](https://deploy-preview-856--i18n-drafts.netlify.app/questions/qa-scripts.en.html)


xfq:
Merging per discussion offline. Copied to CVS as well.

PR #867, opened 2026-04-09
Enhance internationalization explanation

### LINKED ISSUE #785: Internationalization is not only for multilingual applications
We can make it clearer in the [About page](https://www.w3.org/International/i18n-drafts/nav/about) that internationalization is not only useful for multi-locale applications. An internationalized codebase is useful for single-locale applications as well.

xfq:
Fix #785.

Once this is approved, I will update the translation as well.

[Preview](https://deploy-preview-867--i18n-drafts.netlify.app/nav/about.html)

### DIFF
--- a/nav/about.html
+++ b/nav/about.html
@@ -81,7 +81,7 @@ strong { font-weight: 500; }
 
 <p>Translation and localization are NOT what we mean by 'internationalization'. Surprised? Let me explain.</p>
 
-<p>If you internationalize, you design or develop your content, application, specification, and so on, in a way that ensures it will work well for, or can be easily adapted for, users from any culture, region, or language. This is where you address the first set of barriers: not the fact that your user can't read or relate to your product, but the barriers that make it difficult to adapt your product so that they can.</p>
+<p>If you internationalize, you design or develop your content, application, specification, and so on, in a way that ensures it will work well for, or can be easily adapted for, users from any culture, region, or language. This is where you address the first set of barriers: not the fact that your user can't read or relate to your product, but the barriers that make it difficult to <em>adapt</em> your product so that they can.</p>
 
 <p>It's essentially a Quality approach: one that sees you taking action early in the development cycle so that you avoid costly and sometimes prohibitive obstacles when it comes to rolling out your product to new marketplaces.</p>
 
--- a/nav/about.html
+++ b/nav/about.html
@@ -21,7 +21,7 @@ f.status = 'notreviewed';  // should be one of draft, review, published, notrevi
 f.path = '' // what you need to prepend to a URL to get to the /International directory 
 
 // AUTHORS AND TRANSLATORS should fill in these assignments:
-f.thisVersion = { date:'2026-01-15', time:'21:00' } // date and time of latest edits to this document/translation
+f.thisVersion = { date:'2026-04-09', time:'15:41' } // date and time of latest edits to this document/translation
 f.contributors = ''; // people providing useful contributions or feedback during review or at other times
 // make sure to translate the information in the meta description element
 // also make sure that the lang attribute on the html tag is correct!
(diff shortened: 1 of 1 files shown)


r12a:
I think these edits break down some of the original flow.  I propose that, instead, we add the following text as a fourth paragraph.

> Even a single-locale application benefits from internationalization, because it helps you build a code base that handles text, data, formats, and user expectations robustly from the start. It also makes it much less painful to extend your product or application for new markets or users, when the need arises.

nav/about.html
```diff
@@ -81,9 +81,9 @@ <h2>What is Internationalization?</h2>
 
 <p>Translation and localization are NOT what we mean by 'internationalization'. Surprised? Let me explain.</p>
 
-<p>If you internationalize, you design or develop your content, application, specification, and so on, in a way that ensures it will work well for, or can be easily adapted for, users from any culture, region, or language. This is where you address the first set of barriers: not the fact that your user can't read or relate to your product, but the barriers that make it difficult to <em>adapt</em> your product so that they can.</p>
+<p>If you internationalize, you design or develop your content, application, specification, and so on, in a way that ensures it will work well for users, and can be easily adapted for users, from any culture, region, or language. This is not useful only for products that already support multiple locales. Even a single-locale application benefits from internationalization, because it helps you build a code base that handles text, data, formats, and user expectations robustly from the start.</p>
```
aphillips: No comma after "adapted for users"

"This is not useful only" => "This is not only useful"

aphillips (changes requested):
English grammatical/stylistic nits

nav/about.html
r12a: @aphillips i think you're commenting on something that Fuqiao and i agreed to change earlier today.  See my comment about leaving the text as is, and adding a fourth paragraph.

nav/about.html
```diff
@@ -81,9 +81,9 @@ <h2>What is Internationalization?</h2>
...
 
-<p>It's essentially a Quality approach: one that sees you taking action early in the development cycle so that you avoid costly and sometimes prohibitive obstacles when it comes to rolling out your product to new marketplaces.</p>
+<p>It's essentially a Quality approach: one that sees you addressing the first set of barriers early in the development cycle: not the fact that your user can't read or relate to your product, but the barriers that make it difficult to <em>adapt</em> your product so that they can. That early work also helps you avoid costly and sometimes prohibitive obstacles when it comes to rolling out your product to new marketplaces.</p>
```
jsahleen: Quality should not be capitalized, and there are two colons in this sentence, which is odd and difficult to understand.

Suggest:

"Internationalization is essentially a quality approach that addresses a core set of barriers early in the development cycle. It makes it possible to <em>adapt</em> your product while maintaining a single code base, so that users from different cultures, regions, and languages can all read and relate to it. Doing internationalization early helps you avoid costly and sometime prohibitive obstacles when it comes to rolling out your product to new marketplaces."

jsahleen (changes requested):
Wording of the final paragraph is a little difficult to parse. I suggested an alternative, but please modify as you see fit.

nav/about.html
r12a: @jsahleen Quality with a capital refers to an industry process/approach, rather than just an adjective.  We used to use that when i worked in Quality engineering at Xerox and my previous organisation. Maybe it doesn't mean that any more, but if it does, it's a neat way of getting the idea across without a large number of words.

Also, see the decision to insert a new paragraph that Fuqiao and i agreed on this morning. It may change the need for your rewrite.  I'd like to keep the original 'Quality' paragraph short and punchy.  (It was part of my original text.)

r12a: @xfq  maybe you could make the change we discussed quickly so that we're all talking about the same thing.

r12a:
Noting that @xfq is probably no longer around, i'll see if i can make the edit myself.

r12a:
@aphillips @jsahleen I reverted the text to the original content, and added the new content in a fourth paragraph just above "A universal code base".  Please comment on that paragraph.

nav/about.html
jsahleen: I would suggest using something like Quality Management, since it's not obvious that this is a named approach and not just a typo.

I'm fine with the original text + the new paragraph. It just didn't read correctly in the modified version.

r12a: >  It just didn't read correctly in the modified version.

Yeah, i agree. That's why i suggested to Fuqiao that we don't try to blend stuff in, but add a new para.

jsahleen (approved):

PR #868, opened 2026-04-09
Update inline-space with new Chinese text examples

xfq:


r12a:
Thanks much for this (very nice) bit of text !

The PR replaces the wrong place, so i'll just close it, but you can see the 'text' in place as an image at https://w3c.github.io/i18n-drafts/articles/styling/inline-space.en#punc_chinese

PR #875, opened 2026-04-14
Enhance article with historical context on message formatting

### LINKED ISSUE #846: [articles/composite-messages/index]  Refine the plural/gener sections
[source] (https://www.w3.org/International/articles/composite-messages/index.en.html) [en]

I think the plural/gender sections would be more useful if they also covered whole-message localization, plural handling, or message-format patterns, instead of mostly steering readers toward topic-comment phrasing.

xfq:
Fix #846.

I’ve made relatively minor updates. In the future, I might consider writing a dedicated article specifically about MF2 (see also https://github.com/w3c/i18n-drafts/issues/888 ).

### DIFF
--- a/articles/composite-messages/index.en.html
+++ b/articles/composite-messages/index.en.html
@@ -74,6 +74,8 @@ script developers (PHP, JSP, etc.), XHTML/HTML coders (using editors or scriptin
 
 <p>After reviewing the concepts and issues relating to text fragmentation and string reuse, we will look at what works and what doesn't.</p>
 
+<p>This article was originally written in a tooling environment that offered fewer ways to handle plural-sensitive, ordinal-sensitive, and other selectable messages. The problems described here still matter, but some of the recommendations should be read in that historical context. Where modern message-formatting systems are available, it is often preferable to localize complete messages and let the message format handle the necessary selection logic. Topic-comment (also called subject: predicate) remains useful as a workaround when such support is unavailable or impractical.</p>
+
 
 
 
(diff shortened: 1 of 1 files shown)


articles/composite-messages/index.en.html
```diff
@@ -74,6 +74,8 @@ <h1>Working with Composite Messages</h1>
 
 <p>After reviewing the concepts and issues relating to text fragmentation and string reuse, we will look at what works and what doesn't.</p>
 
+<p>This article was originally written in a tooling environment that offered fewer ways to handle plural-sensitive, ordinal-sensitive, and other selectable messages. The problems described here still matter, but some of the recommendations should be read in that historical context. Where modern message-formatting systems are available, it is often preferable to localize complete messages and let the message format handle the necessary selection logic. Topic-comment (also called subject: predicate) remains useful as a workaround when such support is unavailable or impractical.</p>
```
eemeli: "Selectable" is a bit of an odd term to use here. I'd call the generic class of messages with multiple variants as "multivariant", but it's probably better to avoid the term altogether.
```suggestion
<p>This article was originally written in a tooling environment that offered fewer ways to handle plural-sensitive, ordinal-sensitive, and other messages with multiple variants. The problems described here still matter, but some of the recommendations should be read in that historical context. Where modern message-formatting systems are available, it is often preferable to localize complete messages and let the message format handle the necessary selection logic. Topic-comment (also called subject: predicate) remains useful as a workaround when such support is unavailable or impractical.</p>
```

aphillips (commented):
Changes, including @eemeli's suggestion, look okay to me.

jsahleen (approved):
LGTM

articles/composite-messages/index.en.html
xfq: Thank you! Fixed.

PR #877, opened 2026-04-14
Change language enablement page to redirect to new location

### LINKED ISSUE #869: redirect old page
https://www.w3.org/International/i18n-drafts/getting-started/languagedev_participation.html should probably redirect to https://www.w3.org/International/i18n-drafts/pages/languagedev_participation

xfq:
Fix #869.


xfq:
Merging after discussing with @r12a.

PR #878, opened 2026-04-17
Add metadata documentation for translations.js

xfq:


r12a:
This looks very good!  However, i think we need to either link to it from or add it to the editorial guidelines at https://www.w3.org/International/i18n-activity/guidelines/editing

If not including in those guidelines, i think we need to link to it from https://github.com/w3c/i18n-editors/?tab=readme-ov-file, which is where we should be able to find all editor-related instructions and descriptions.

I'd be inclined to also describe the validation stuff or link to it from the i18n-editors page.

I wasn't able to find any other documentation of the translations.js file that i might have done (though i do think there may be something somewhere — never mind, your description in this PR probably does everything that's needed).

xfq:
Thanks, I'll add links in https://www.w3.org/International/i18n-activity/guidelines/editing and i18n-editors.

PR #884, opened 2026-04-20
Add initial draft for input method article

### LINKED ISSUE #879: An article about handling input methods in web development
When building a website, it is easy to encounter issues related to the improper handling of IME. We can consider writing an article on this topic.

As an example, in chat boxes, search fields, comment forms, and any web control where Enter or Return triggers submission, sending, saving, or dialog acceptance, when a CJK user presses Enter to confirm a candidate selection during IME composition, some applications incorrectly send the message before the user has finished entering text.

Web platform has a composition-related state: KeyboardEvent.isComposing. Applications that bypass these platform semantics with raw Enter handlers often fail. Web apps should never execute a send/submit/save action on Enter while composition is active.

UI component libraries and frameworks should treat IME-safe submission as a default contract. Ideally, a reusable “message input” component should be safe out of the box in all browsers and there should be tests about it in wpt.

This is just one example. There are other issues related to input methods. Some may be browser bugs, some may be due to the need for improvement in spec design, and others may require educating better developers to take input methods into account.

Related: https://github.com/w3c/jlreq/issues/484

xfq:
Fix #879.

[Preview](https://deploy-preview-884--i18n-drafts.netlify.app/articles/input-method/index.en.html)


r12a (commented):
From an initial scan, some initial suggestions:

1. make the page title more specific to the topic
2. provide a worked example of steps involved when a user types using an IME 
3. 'respect composition state' - i'm not sure the reader will understand what this means

xfq:
I've updated the title and content.

tats-u:
In the first place, isn't "Send by Ctrl + Enter" better especially when you have to carefully craft your message (especially for your important customer or boss, or expensive AI models like Claude Opus and GPT 5.x Pro)?

tats-u:
How to input CJK languages for non-East-Asian people: https://github.com/microsoft/vscode/wiki/IME-Test

xfq:
> In the first place, isn't "Send by Ctrl + Enter" better especially when you have to carefully craft your message (especially for your important customer or boss, or expensive AI models like Claude Opus and GPT 5.x Pro)?

Good suggestion. I adjusted the text.

tats-u:
If Ctrl (or Cmd) was pressed during initial transmission, the application should ask the user if Enter's behavior should be changed to newline because the user prefer the newline-by-Enter behavior.

PR #889, opened 2026-04-23
Add a guide on transliteration, translation, and transcription

xfq:
[Preview](https://github.com/w3c/i18n-drafts/blob/xfq/transliteration-transcription/questions/qa-transliteration-transcription.md)


questions/qa-transliteration-transcription.md
```diff
@@ -0,0 +1,109 @@
...
+
+Transliteration, translation, and transcription solve different problems, and choosing the wrong one can quietly damage usability and trust.
+
+It is important to know when to use each one, how to store the data, and how to design search behavior.
+
+## The three concepts
+
+### Translation
+
+Translation conveys meaning from one language to another. For example, “White House” is translated to “白宫” in Chinese.
```
r12a, line 19: To better exemplify the difference from translit and transc, consider reversing the example, ie. 
```suggestion
Translation conveys meaning from one language to another. For example, “白宫” in Chinese is translated to “White House” in English.
```
An even better approach would be to find a word in a foreign language that has different Latin renderings for translation, transliteration, and transcription.  Then show that example in all three sections.

questions/qa-transliteration-transcription.md
```diff
@@ -0,0 +1,109 @@
...
+
+For names, translation is not always inappropriate. When product teams rely on machine processes or try to localize every string, mistakes sometimes happen. A person named “Grace” should not become the equivalent of “grace” in another language.
```
r12a, line 21: we have an article about the translate attribute, which shows you how to shield words like this.

But you seem to be arguing in the first sentence that translation of names can be ok, but then giving an example of when it's not ok.  Did you mix things up?

articles/transliteration-transcription/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Translation
+
+Translation conveys meaning from one language to another. For example, “White House” is translated to “白宫” in Chinese.
+
+For names, translation is not always inappropriate. When product teams rely on machine processes or try to localize every string, mistakes sometimes happen. A person named “Grace” should not become the equivalent of “grace” in another language.
+
+A city, institution, or title may be different, because some proper nouns do have established translated forms.
+
+Usually, you should translate names only when there is a strong convention or an explicit editorial policy for doing so.
```
r12a: This looks like a comment that should appear at the beginning of this section, followed by the caveats and examples.

articles/transliteration-transcription/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Transliteration
+
+Transliteration maps characters from one script to another, with the goal of preserving the written form as systematically as possible. For example, the Russian word Москва can be transliterated to Moskva.
+
+Transliteration answers the question, “How do I represent this spelling in another script?” It does not primarily aim to preserve pronunciation or meaning.
+
+This matters because multiple transliteration systems may exist for the same source script. The result can vary depending on whether you prioritize linguistic accuracy, passport conventions, library standards, or user familiarity.
```
r12a: transliteration systems -> transcription systems ?

A major difference that is particular to transliteration is that it allows lossless round-trip conversions between orthographies. I think you need to stress that point. Perhaps show by adding to the first paragraph, that Moskva can then be converted back to Москва accurately.

articles/transliteration-transcription/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Transcription
+
+Transcription represents how something sounds.
```
r12a: Strictly speaking, transliteration is a type of transcription.  So here we're talking about 'other types of transcription'.  

Such transcriptions typically attempt to represent the sound using conventions aimed at a particular type of reader.  For accurate phonetic transcription, IPA is a useful convention (i really think you should mention IPA here).  For the layman, transcriptions often try to convey sounds using the conventions of English or whatever language they speak, resulting in things like 'mosskvuh' (yuk!).  But there are also many standard transcription approaches, such as those used for transcribing Indian languages.  For Japanese, in addition to kana, there are, of course, romaji transcriptions.  There are at least 2 main approaches (Wylie and another i forgot the name of).

A key fact to stress about these transcriptions is that they don't usually allow you convert the transcription back to the original orthography.

questions/qa-transliteration-transcription.md
```diff
@@ -0,0 +1,109 @@
...
+* a common Latin-script spelling for travel or payments
+* a pronunciation guide for customer support or speech systems
+
+Treating all of that as one field called “translated name” often leads to errors.
+
+## Identity first
+
+For names, the first principle is simple: preserve the original form whenever possible.
+
+The original script form is often the most authoritative representation of a person, place, or organization. That does not mean it should always be the only visible form. It means it should be stored distinctly and treated as primary data.
```
r12a, line 66: You can also include here examples that show how the native script disambiguates pronunciation and identity better than romanisations.  I think there's an example in the Personal Names article.  For example, there are many han characters representing the sound 'shi'....

questions/qa-transliteration-transcription.md
```diff
@@ -0,0 +1,109 @@
...
+When dealing with names, search, and fallback, ask these questions.
+
+### What is the user trying to do?
+
+* Identify a person or place?
+* Pronounce a name?
+* Read content in an unfamiliar script?
+* Match records across systems?
+
+The answer determines whether you need translation, transliteration, transcription, or some combination.
```
r12a, line 99: By now, i'm beginning to realise that this article is not about Transliteration vs. translation vs. transcription per se, but rather about whether to use Transliteration vs. translation vs. transcription when representing names.  Perhaps you should change the title and introduction so that it's clearer what is the purpose of the article and what the reader will get out of reading it.

questions/qa-transliteration-transcription.md
```diff
@@ -0,0 +1,109 @@
...
+
+Translation preserves meaning, transliteration preserves script-level correspondence, and transcription preserves sound.
+
+In multilingual products, especially around names, search, and fallback, that distinction affects whether users can find things, recognize themselves, trust the interface, and complete tasks without friction.
+
+The most reliable approach is to preserve original forms, model alternative forms explicitly, design search to connect them intelligently, and define fallback behavior on purpose. It will make the product feel more accurate, more respectful, and more usable across languages and scripts.
+
+## Further reading
+
+* [Internationalization techniques: Authoring web pages](https://www.w3.org/International/techniques/authoring-html)
```
r12a, line 121: Proabably also link directly to the Personal Names article.

Definitely you should also link to the LE Index.

r12a (changes requested):

questions/qa-transliteration-transcription.md
xfq: I’ve thought about that too, but I couldn’t find a better example. Do you have any good examples?

questions/qa-transliteration-transcription.md
xfq: I didn't quite understand what you meant; what I said above was “for names, translation is not always inappropriate”.

Sometimes people translate the meaning of personal names directly, but this is not the mainstream practice.

articles/transliteration-transcription/index.md
xfq: fixed

articles/transliteration-transcription/index.md
xfq: I made some adjustments.

articles/transliteration-transcription/index.md
xfq: OK, I made some adjustments.

questions/qa-transliteration-transcription.md
xfq: done

questions/qa-transliteration-transcription.md
xfq: done

questions/qa-transliteration-transcription.md
xfq: done

PR #899, opened 2026-05-10
Enhance directionality attributes in HTML guidance

### LINKED ISSUE #898: `dir` of an image
When the `dir` of an image is set to "rtl", the HTML attributes `title` and `alt` will be right-aligned and flow from right to left. This should probably be mentioned in https://www.w3.org/International/questions/qa-html-dir or https://www.w3.org/International/articles/inline-bidi-markup/

xfq:
Fix #898.

[Preview](https://deploy-preview-899--i18n-drafts.netlify.app/questions/qa-html-dir.en.html)


xfq:

PR #914, opened 2026-07-02
articles/vertical-text   Various updates and improvements.

r12a:
[Article](https://w3c.github.io/i18n-drafts/articles/vertical-text/index.en.html)


r12a:
Taking the PR route here so that changes are visible.  If there is an issue, we can raise separately.  However, i am about to apply the latest article template and fix validation issues, etc, which i'll do with a direct commit.

PR #918, opened 2026-07-08
Add DOM Localization

### LINKED ISSUE #915: Add DOM localization
Should we add a link to DOM localization in this page: https://www.w3.org/International/i18n-drafts/nav/specdev

/cc @eemeli

xfq:
Fix #915.

[Preview](https://deploy-preview-918--i18n-drafts.netlify.app/nav/specdev.html)


eemeli (approved):
Should we also mention our own work on [message resources](https://github.com/w3c/i18n-discuss/blob/gh-pages/explainers/message-resources.md)?

PR #920, opened 2026-07-08
Use real translation URLs with progressive enhancement for the cookie prompt

### LINKED ISSUE #919: Use real translation URLs with progressive enhancement for the cookie prompt
Language links in our articles use `href="#"` plus inline `onclick` at `article-2022.js`. See https://www.w3.org/International/questions/qa-lang-why.en.html as an example.

We should use real translation URLs with progressive enhancement for the cookie prompt, so that navigation is no longer dependent on JavaScript.

With real `href`s, the language links work correctly for no-JS users, copied links, status-bar previews, middle-click/open-in-new-tab, search/crawlers, and assistive tech. The browser sees them as actual links instead of fake `#` anchors.

The cookie prompt becomes progressive enhancement: if JS runs, the click handler asks about setting it. Either way, normal navigation continues to the real translation URL. If JS fails or is disabled, the user still reaches the translation.

xfq:
Fix #919.

Example page after the change: https://deploy-preview-920--i18n-drafts.netlify.app/questions/qa-lang-why.en.html

Before the change: https://www.w3.org/International/questions/qa-lang-why.en.html

### DIFF
--- a/javascript/doc-structure/article-2022.js
+++ b/javascript/doc-structure/article-2022.js
@@ -56,11 +56,24 @@ function showExample (path) {
 	}
 
 
-function setLanguagePreference (targetLang) {
+function getLanguagePreferenceMessage (cLang, targetLang) {
+	var messages = {}
+	if (typeof cn !== 'undefined') messages = cn
+	var cMsg = messages[cLang]
+	var targetMsg = messages[targetLang]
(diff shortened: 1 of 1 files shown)


xfq:
I’ve merged the changes, uploaded them to CVS, and tested them. Everything seems fine.

PR #930, opened 2026-07-13
Fix typos in ruby styling

xfq:


r12a:
Thanks for this!  I have made the changes locally, and they'll appear in my next commit.

PR #937, opened 2026-08-06
Add more results & explanations for text-spacing-trim

r12a:
[Preview](https://deploy-preview-937--i18n-drafts.netlify.app/articles/styling/inline-space.en.html#punctuation_initial)


xfq:
Thanks. I just updated the Chinese translation.

PR #945, opened 2026-08-13
Replace Jukuji to Jukujikun

himorin:
From discussion in [JL-TF call](https://github.com/w3c/jlreq/discussions/488) (minutes under preparation), in Japanese Jukuji (熟字) and Jukujikun (熟字訓) (in theory) point different set of sequence of Kanji character, Jukuji shows special meaning from composed sequence of Kanji character, and Jukujikun has special Kunnyomi over entire Kanji characters which is different from combination of Kunnyomi for each Kanji character - so Jukuji would be larger set than Jukujikun.
No concrete agreement has been secured whether `Jukujikun` (alone) or `Jukujikun annotation` fits well for ruby text.

/cc @kidayasuo @KobayashiToshi


r12a:
We're using the term 'jukuji' in a way that has been used for several years now.  If no concrete agreement has been reached about this, no action is appropriate.  I will close the PR.  It can be reopened later, if appropriate.

himorin:
@r12a noted. thank you for official record on this.

kidayasuo:
Comment from the JLReq TF meeting on 2026-9-1.

Either "jukuji reading" or "jukuji-kun" is a correct term for phonetic readings and we do not have a preference one way or the other. Since "jukuji reading" is already used in JLReq, it would be reasonable to keep using it.

We did find one place where the term is used incorrectly. Item 5 of the following document reads "Jukuji are phonetic readings where…" It should read "Jukuji **reading** are phonetic readings……".
https://www.w3.org/International/datespace/2010/02/jlreq-examples/

Thank you.

